PR #22, opened 2019-08-16
Add one option to disable certificate verification.

yong-li:


include/libnuraft/asio_service_options.hxx
```diff
@@ -66,6 +66,7 @@ struct asio_service_options {
     asio_service_options()
         : thread_pool_size_(0)
         , enable_ssl_(false)
+        , verify_peer_(true)
```
greensky00: Could you please change it to `skip_verification_(false)`? That seems to me better.

yong-li: sure.

PR #25, opened 2019-08-19
Expose restart election timer. 

yong-li:
The idea is the following.  For the provision workflow, the following would happen:
1. Every node is started with `skip_initial_election_timeout_ = true`.
2. The agent will pick one node, and call `/initialize` on this node.  At this time `restart_election_timer()` will be called to elect this node to be the leader.

The idea is to make `/startserver` consistent among all nodes, and use the `/initialize` call to make the designated node the leader.

### DIFF
--- a/src/handle_timeout.cxx
+++ b/src/handle_timeout.cxx
@@ -102,7 +102,8 @@ void raft_server::handle_hb_timeout(int32 srv_id) {
 
 void raft_server::restart_election_timer() {
     // don't start the election timer while this server is still catching up the logs
-    if (catching_up_) {
+    // or this server is the leader
+    if (catching_up_ || role_ == srv_role::leader) {
         return;
     }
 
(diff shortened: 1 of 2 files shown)


src/handle_timeout.cxx
```diff
@@ -102,7 +102,8 @@ void raft_server::handle_hb_timeout(int32 srv_id) {
 
 void raft_server::restart_election_timer() {
     // don't start the election timer while this server is still catching up the logs
-    if (catching_up_) {
+    // or this server is the leader
+    if (catching_up_ || role_ == srv_role::leader) {
```
greensky00: Would you please add `recur_lock(lock_);` at the beginning of this function? So far this function was internal use only, which means all callers already acquire `lock_` before reach here. But now we expose it to users so that this function should protect itself.

PR #33, opened 2019-08-27
Reduce target priority upon term increment

greensky00:
* Previously we decay target priority on election timeout, but that
is not sufficient if there is an aggressive node which sends vote
request continuously. Then voters will reset their timer and never
face election timeout, thus we end up with endless leader election.

* To avoid such issue, we reduce the target priority on term increment
as well. Term increment when there is no heartbeat most likely implies
no leader has been elected in previous term.

* Added test case.

### DIFF
--- a/src/raft_server.cxx
+++ b/src/raft_server.cxx
@@ -455,7 +455,18 @@ ptr<resp_msg> raft_server::process_req(req_msg& req) {
          req.get_type() == msg_type::request_vote_request ||
          req.get_type() == msg_type::install_snapshot_request ) {
         // we allow the server to be continue after term updated to save a round message
-        update_term(req.get_term());
+        bool term_updated = update_term(req.get_term());
+
+        if ( !im_learner_ &&
+             !hb_alive_ &&
+             term_updated &&
+             req.get_type() == msg_type::request_vote_request ) {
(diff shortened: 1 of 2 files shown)


src/raft_server.cxx
```diff
@@ -455,7 +455,18 @@ ptr<resp_msg> raft_server::process_req(req_msg& req) {
...
         // we allow the server to be continue after term updated to save a round message
-        update_term(req.get_term());
+        bool term_updated = update_term(req.get_term());
+
+        if ( !im_learner_ &&
+             !hb_alive_ &&
+             term_updated &&
+             req.get_type() == msg_type::request_vote_request ) {
+            // If someone has newer term, that means leader has not been
+            // elected in the current term, and this node's election timer
```
andy-yx-chen: this is misleading, a newer term does not mean leader not elected in the current term, think about this, a follower did not receive hb for quite some time and initiate a new term, there could be a leader in previous term

greensky00: Thanks Andy. We have pre-vote so that follower cannot increase its term without the agreement of quorum. The increment of term only happens when more than majority of servers don't receive heartbeat from the (possibly previous) leader.

PR #35, opened 2019-08-29
Fix compile errors on Mac

greensky00:
* Need to link dependencies for shared library.


tisonkun:
@greensky00 shall we do the same for static_lib? I use:

```
target_link_libraries(mephisto PRIVATE static_lib)
```

But when linking, it fails with SSL symbols not found.

Also, `${LIBRARIES}` var is not usable in this scope.

PR #39, opened 2019-09-09
Fix follower will abort after a while when raft log of follower is out of range

shangshengshi:


src/handle_append_entries.cxx
```diff
@@ -338,6 +338,8 @@ ptr<resp_msg> raft_server::handle_append_entries(req_msg& req)
         }
     }
 
+    leader_ = req.get_src();
```
greensky00: We should not change `leader_` here before checking whether the given message is valid.

greensky00:
Closing it as we will come up with other approach.

PR #87, opened 2019-12-27
Set thread name for raft threads for Mac OS

shigui1989:


greensky00:
Thanks @shigui1989 

Just FYI, while I was testing your PR, I found one test failure:
```
[ .... ] uncommitted config for new leader test
   === TEST MESSAGE (BEGIN) ===

        time: 2019-12-27 12:07:54.145422
      thread: e624
          in: uncommitted_conf_new_leader_test()
          at: /Users/junahn/sandbox/raft/tests/unit/failure_test.cxx:449
    value of: c_config->get_server(2).get()
    expected: NULL
      actual: 0x7fb982d015a8
[ FAIL ] uncommitted config for new leader test (1.0 s)
```
but noticed that it is not a regression by your change. I made a separate fix: https://github.com/eBay/NuRaft/pull/89

PR #93, opened 2020-01-13
Update sm_commit_index_ after state_machine commit is called.

shigui1989:


include/libnuraft/raft_server.hxx
```diff
@@ -572,8 +572,8 @@ protected:
     void commit_in_bg();
     void append_entries_in_bg();
 
-    void commit_app_log(ptr<log_entry>& le, bool need_to_handle_commit_elem);
-    void commit_conf(ptr<log_entry>& le);
+    void commit_app_log(ptr<log_entry>& le, bool need_to_handle_commit_elem, ulong sm_committing_index);
```
greensky00: Please don't exceed 90 chars per line.

src/handle_commit.cxx
```diff
@@ -130,29 +130,30 @@ void raft_server::commit_in_bg() {
...
             if (le->get_val_type() == log_val_type::app_log) {
-                commit_app_log(le, need_to_handle_commit_elem);
+                commit_app_log(le, need_to_handle_commit_elem, sm_committing_index);
 
             } else if (le->get_val_type() == log_val_type::conf) {
-                commit_conf(le);
+                commit_conf(le, sm_committing_index);
             }
 
+            sm_commit_index_ = sm_committing_index;
```
greensky00: To be safe, we should use compare-and-swap here.

PR #128, opened 2020-08-04
Make the hint size which is the preferred size of the next batch of l…

shigui1989:
…ogs to be sent from the leader can be negative value

@greensky00 See detailed info from MONSTOR-13421

### DIFF
--- a/include/libnuraft/state_machine.hxx
+++ b/include/libnuraft/state_machine.hxx
@@ -130,7 +130,7 @@ public:
      *         `positive value` indicates at least one log can be sent,
      *         (the size of that log may be bigger than this hint size).
      *         `negative value` indicates no log should be sent since this
-     *          follower is busy handling pending logs.
+     *         follower is busy handling pending logs.
      */
     virtual int64 get_next_batch_size_hint_in_bytes() { return 0; }
 
--- a/include/libnuraft/state_machine.hxx
+++ b/include/libnuraft/state_machine.hxx
@@ -125,7 +125,7 @@ public:
      *
      * Only applicable on followers.
      *
-     * @return the preferred size of the next log batch
+     * @return The preferred size of the next log batch.
      *         `0` indicates no preferred size (any size is good).
      *         `positive value` indicates at least one log can be sent,
      *         (the size of that log may be bigger than this hint size).
--- a/include/libnuraft/state_machine.hxx
+++ b/include/libnuraft/state_machine.hxx
@@ -123,13 +123,16 @@ public:
      * Return a hint about the preferred size (in number of bytes)
      * of the next batch of logs to be sent from the leader.
      *
-     * Return 0 to indicate no preferred size (any size is good).
-     *
      * Only applicable on followers.
      *
      * @return the preferred size of the next log batch
+     *         `0` indicates no preferred size (any size is good).
+     *         `positive value` indicates at least one log can be sent,
+     *         (the size of that log may be bigger than this hint size).
+     *         `negative value` indicates no log should be sent since this
+     *          follower is busy handling pending logs.
      */
-    virtual ulong get_next_batch_size_hint_in_bytes() { return 0; }
+    virtual int64 get_next_batch_size_hint_in_bytes() { return 0; }
 
     /**
      * (Deprecated)
--- a/src/asio_service.cxx
+++ b/src/asio_service.cxx
@@ -574,7 +574,7 @@ private:
             // Hint is given, set the flag.
             flags |= INCLUDE_HINT;
             // For future extension, we will put 2-byte version and 2-byte length.
-            resp_hint_size += sizeof(uint16_t) * 2 + sizeof(ulong);
+            resp_hint_size += sizeof(uint16_t) * 2 + sizeof(int64);
         }
 
         size_t carried_data_size = resp_meta_size + resp_hint_size + resp_ctx_size;
@@ -610,7 +610,7 @@ private:
             const uint16_t CUR_HINT_VERSION = 0;
             bs.put_u16(CUR_HINT_VERSION);
             bs.put_u16(sizeof(ulong));
-            bs.put_u64(resp->get_next_batch_size_hint_in_bytes());
+            bs.put_i64(resp->get_next_batch_size_hint_in_bytes());
         }
 
         if (resp_ctx_size) {
@@ -1377,7 +1377,7 @@ private:
             uint16_t hint_version = bs.get_u16();
             (void)hint_version;
             hint_len = bs.get_u16();
-            rsp->set_next_batch_size_hint_in_bytes(bs.get_u64());
+            rsp->set_next_batch_size_hint_in_bytes(bs.get_i64());
             remaining_len -= sizeof(uint16_t) * 2 + hint_len;
         }
 
--- a/tests/unit/asio_service_test.cxx
+++ b/tests/unit/asio_service_test.cxx
@@ -644,7 +644,9 @@ int response_hint_test(bool with_meta) {
         ee->getTestSm()->set_next_batch_size_hint_in_bytes(-1);
     }
 
-    for (size_t ii=0; ii<NUM; ++ii) {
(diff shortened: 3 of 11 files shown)


src/asio_service.cxx
```diff
@@ -610,7 +610,7 @@ class rpc_session : public std::enable_shared_from_this<rpc_session> {
             const uint16_t CUR_HINT_VERSION = 0;
             bs.put_u16(CUR_HINT_VERSION);
             bs.put_u16(sizeof(ulong));
-            bs.put_u64(resp->get_next_batch_size_hint_in_bytes());
+            bs.put_i64(resp->get_next_batch_size_hint_in_bytes());
```
greensky00, line 613: During a rolling upgrade, old Raft instance with `u64` and new Raft instance with `i64` will be running in the same group. Will it be safe? If `i64` server returns a negative number, it is interpreted as a big number for `u64` server, and the only impact I can imagine is that there is no throttling, right?

tests/unit/asio_service_test.cxx
```diff
@@ -639,13 +639,33 @@ int response_hint_test(bool with_meta) {
     CHK_OK( s2.getTestSm()->isSame( *s1.getTestSm() ) );
     CHK_OK( s3.getTestSm()->isSame( *s1.getTestSm() ) );
 
+    _msg("enable batch size hint with negative value\n");
+    for (RaftAsioPkg* ee: pkgs) {
+        ee->getTestSm()->set_next_batch_size_hint_in_bytes(-1);
+    }
+
+    // on Mac, due to sync append, leader reelect will happen
+    // on Ubuntu, due to sync append, 100 times append are very slow
```
greensky00, line 648: Please adjust `client_req_timeout_` to a shorter number for this test, so as to make it faster, using `update_params` API. The current default is 10 seconds. 
https://github.com/eBay/NuRaft/blob/6d371ddcc6bc0a9b8aaad4e57e9859adbc111d85/tests/unit/raft_package_asio.hxx#L123

greensky00: Also may need to adjust `leadership_expiry_` to `-1`, to avoid unwanted leader election.

src/asio_service.cxx
shigui1989: Yes. I think the only impact is that `hint size` will not take effect. Besides, in monstordb, we still have 16M restriction. So the whole throttling should still be there.

tests/unit/asio_service_test.cxx
shigui1989: update `client_req_timeout_` is done.
About adjusting `leadership_expiry_ `, it may be not useful. On Mac, the re-elect is caused by the follower did not receive the heartbeat and is triggered by follower, not by leader.

Besides, with negative hint size, raft only can not send append entry request with raft logs, but the empty append entry request still can work as a heartbeat.

Since we only do client request on raft server 1, so if leader changes, then the following client request on old leader may fail quickly. For this situation, the purpose of this test case may can not be achieved, but it should not influence the successful finish of this test case.

include/libnuraft/state_machine.hxx
```diff
@@ -123,13 +123,16 @@ public:
...
-     * Return 0 to indicate no preferred size (any size is good).
-     *
      * Only applicable on followers.
      *
      * @return the preferred size of the next log batch
+     *         `0` indicates no preferred size (any size is good).
+     *         `positive value` indicates at least one log can be sent,
+     *         (the size of that log may be bigger than this hint size).
+     *         `negative value` indicates no log should be sent since this
+     *          follower is busy handling pending logs.
```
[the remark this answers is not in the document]
shigui1989: Done

tests/unit/asio_service_test.cxx
```diff
@@ -657,6 +657,32 @@ int response_hint_test(bool with_meta) {
...
+    s1.raftServer->update_params(params);
+
+    for (size_t ii=0; ii<3; ++ii) {
+        std::string msg_str = "3rd_" + std::to_string(ii);
+        ptr<buffer> msg = buffer::alloc(sizeof(uint32_t) + msg_str.size());
+        buffer_serializer bs(msg);
+        bs.put_str(msg_str);
+        s1.raftServer->append_entries( {msg} );
+    }
+    TestSuite::sleep_sec(1, "wait for replication but actually no replication happen");
```
greensky00, line 680: If you temporarily increase this sleep time much longer, you can see that Raft is overwhelming CPU, because `need_to_catchup` is `true`. The leader is continuously sending meaningless messages to followers whose hint is `-1`. I think if hint is `-1`, `need_to_catchup` flag should be `false` so that the leader sends heartbeats only.
```
2020-08-05T12:22:34.782_025-07:00 [3b34] [DEBG] Receive a append_entries_response message from peer 3 with Result=1, Term=1, NextIndex=204  [raft_server.cxx:681, handle_peer_resp()]
2020-08-05T12:22:34.782_029-07:00 [3b34] [TRAC] src: 3, dst: 1, resp->get_term(): 1 [raft_server.cxx:684, handle_peer_resp()]
2020-08-05T12:22:34.782_034-07:00 [3b34] [TRAC] handle append entries resp (from 3), resp.get_next_idx(): 204   [handle_append_entries.cxx:800, handle_append_entries_resp()]
2020-08-05T12:22:34.782_038-07:00 [3b34] [TRAC] peer 3 batch size hint: -1 bytes    [handle_append_entries.cxx:803, handle_append_entries_resp()]
2020-08-05T12:22:34.782_041-07:00 [3b34] [TRAC] peer 3, prev matched idx: 203, new matched idx: 203 [handle_append_entries.cxx:815, handle_append_entries_resp()]
2020-08-05T12:22:34.782_049-07:00 [3b34] [TRAC] quorum idx 1, 206 203 203   [handle_append_entries.cxx:956, get_expected_committed_log_idx()]
2020-08-05T12:22:34.782_054-07:00 [3b34] [TRAC] local log idx 206, target_commit_idx 203, quick_commit_index_ 203, state_->get_commit_idx() 203 [handle_commit.cxx:59, commit()]
2020-08-05T12:22:34.782_057-07:00 [3b34] [DEBG] reqeust append entries need to catchup, p 3 [handle_append_entries.cxx:924, handle_append_entries_resp()]
2020-08-05T12:22:34.782_062-07:00 [3b34] [TRAC] send request to 3   [handle_append_entries.cxx:154, request_append_entries()]
2020-08-05T12:22:34.782_066-07:00 [3b34] [TRAC] last_log_idx: 203, starting_idx: 191, cur_nxt_idx: 207  [handle_append_entries.cxx:288, create_append_entries_req()]
2020-08-05T12:22:34.782_069-07:00 [3b34] [DEBG] last sent log (204) to peer 3 is not applied, cnt 50843 [handle_append_entries.cxx:308, create_append_entries_req()]
2020-08-05T12:22:34.782_071-07:00 [3b34] [DEBG] reduce end_idx 207 -> 205   [handle_append_entries.cxx:312, create_append_entries_req()]
2020-08-05T12:22:34.782_077-07:00 [3b34] [TRAC] adjusted end_idx due to batch size hint: 205 -> 204 [handle_append_entries.cxx:386, create_append_entries_req()]
2020-08-05T12:22:34.782_081-07:00 [3b34] [DEBG] append_entries for 3 with LastLogIndex=203, LastLogTerm=1, EntriesLength=0, CommitIndex=203, Term=1, peer_last_sent_idx 204 [handle_append_entries.cxx:394, create_append_entries_req()]
2020-08-05T12:22:34.782_084-07:00 [3b34] [TRAC] EMPTY PAYLOAD   [handle_append_entries.cxx:396, create_append_entries_req()]
2020-08-05T12:22:34.782_092-07:00 [3b34] [TRAC] send req 1 -> 3, type append_entries_request    [peer.cxx:43, send_req()]
2020-08-05T12:22:34.782_123-07:00 [3b34] [TRAC] sent    [handle_append_entries.cxx:224, request_append_entries()]
```

greensky00: I fixed it by myself and looks good. Let me merge it thanks.

PR #133, opened 2020-09-04
Auto-forwarding now waits for a result when using blocking mode before returning

FletcherJ1:
I've added the wait method to the cmd_result class because calling get won't work due to how the presult variable is instantiated. 
https://github.com/eBay/NuRaft/blob/17aab8d633b0187dc4a5d862aa75faf3ef91f5cd/src/handle_user_cmd.cxx#L168-L170
When presult is made, it calls a constructor that sets has_result_ to true. Then, if get() was used it would see that has_result_ is true and return immediately and not wait for the set_result call. Therefore, the explicit wait method will guarantee the current thread to wait until a call to set_result happens.

### DIFF
--- a/include/libnuraft/async.hxx
+++ b/include/libnuraft/async.hxx
@@ -268,6 +268,16 @@ public:
         return empty_result_;
     }
 
+    /**
+     * Blocks until set_result is called.
+     *
+     * @return void.
+     */
+    void wait() {
+        std::unique_lock<std::mutex> lock(lock_);
+        cv_.wait(lock);
+    }
+
 private:
     T empty_result_;
     T result_;
--- a/include/libnuraft/async.hxx
+++ b/include/libnuraft/async.hxx
@@ -268,16 +268,6 @@ public:
         return empty_result_;
     }
 
-    /**
-     * Blocks until set_result is called.
-     *
-     * @return void.
-     */
-    void wait() {
-        std::unique_lock<std::mutex> lock(lock_);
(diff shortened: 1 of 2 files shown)


greensky00:
@FletcherJ1 
Thanks for submitting the new PR.

I get your point regarding `has_result_` value. But my understanding is that the code that you are referring to will call the default constructor (i.e., `cmd_result()`) as there is no parameter given for `cmd_result`: `cs_new< cmd_result< ptr<buffer> > >()`

I just checked with GDB and saw that `has_result_` value is `false`.
```
Thread 1 "asio_service_te" hit Breakpoint 1, nuraft::raft_server::send_msg_to_leader (this=0x555555c65bf0,
    req=std::shared_ptr<nuraft::req_msg> (use count 1, weak count 0) = {...})
    at /home/junahn/sandbox/nuraft/src/handle_user_cmd.cxx:187
187	    rpc_cli->send(req, handler);
(gdb) p presult->has_result_
$1 = false
(gdb) p presult->result_
$2 = std::shared_ptr<nuraft::buffer> (empty) = {get() = 0x0}
(gdb)
```

If that's the case, we can just use the existing `get()` function. Could you please double-check? Thanks!

FletcherJ1:
Ah yes, you're correct. I've swapped to using get and it works!

greensky00 (approved):

greensky00:
Merged, thanks!

PR #137, opened 2020-09-15
Hide `worker_handle` and move it to cxx file

greensky00:
* `worker_handle` does not need to be exposed.

### DIFF
--- a/include/libnuraft/global_mgr.hxx
+++ b/include/libnuraft/global_mgr.hxx
@@ -20,7 +20,6 @@ limitations under the License.
 #pragma once
 
 #include "basic_types.hxx"
-#include "event_awaiter.h"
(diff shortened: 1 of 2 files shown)


szmyd:
I think worker_handle needs to be a forward declared pointer not a reference for this to work right?

greensky00:
@szmyd Sorry I don't get your point, `global_mgr` API uses `ptr<worker_handle>` only so that the forward declaration of `struct worker_handle` is enough. Isn't it?

szmyd:
Oh yes, sorry. I'm not used to seeing a forward declare inside a class definition, I misread it as a member declaration. This looks good.

szmyd (approved):

greensky00:
Ok, I'm merging it, thanks!

PR #146, opened 2020-10-20
add build to .gitignore and  enclose typedefs with namespace in basic_type.hxx

sheepgrass:
- add build to .gitignore
- enclose typedefs with namespace in basic_type.hxx


greensky00 (approved):
Thanks @sheepgrass

PR #147, opened 2020-10-20
support setting raft server init options and raft callback in raft_launcher::init()

sheepgrass:
support setting raft server init options and raft callback in raft_launcher::init()

### DIFF
--- a/include/libnuraft/launcher.hxx
+++ b/include/libnuraft/launcher.hxx
@@ -37,6 +37,8 @@ public:
      * @param port_number Port number.
      * @param asio_options ASIO options.
      * @param params Raft parameters.
+     * @param opt Raft server init options.
+     * @param raft_callback Callback function for hooking the operation.
      * @return Raft server instance.
      *         `nullptr` on any errors.
      */
@@ -45,7 +47,9 @@ public:
                           ptr<logger> lg,
                           int port_number,
                           const asio_service::options& asio_options,
-                          const raft_params& params);
+                          const raft_params& params,
+                          const raft_server::init_options opt = raft_server::init_options(),
+                          cb_func::func_type raft_callback = nullptr);
 
     /**
      * Shutdown Raft server and ASIO service.
--- a/include/libnuraft/launcher.hxx
+++ b/include/libnuraft/launcher.hxx
@@ -48,8 +48,7 @@ public:
                           int port_number,
                           const asio_service::options& asio_options,
                           const raft_params& params,
-                          const raft_server::init_options opt = raft_server::init_options(),
-                          cb_func::func_type raft_callback = nullptr);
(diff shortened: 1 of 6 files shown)


include/libnuraft/launcher.hxx
```diff
@@ -45,7 +47,9 @@ public:
                           ptr<logger> lg,
                           int port_number,
                           const asio_service::options& asio_options,
-                          const raft_params& params);
+                          const raft_params& params,
+                          const raft_server::init_options opt = raft_server::init_options(),
```
greensky00, line 51: Could you please use reference?
```C++
const raft_server::init_options& opt = raft_server::init_options()
```

src/launcher.cxx
```diff
@@ -32,7 +32,9 @@ ptr<raft_server> raft_launcher::init(ptr<state_machine> sm,
                                      ptr<logger> lg,
                                      int port_number,
                                      const asio_service::options& asio_options,
-                                     const raft_params& params_given)
+                                     const raft_params& params_given,
+                                     const raft_server::init_options opt,
+                                     cb_func::func_type raft_callback)
```
greensky00, line 37: As I mentioned in the previous PR, I'd like to suggest you to move `cb_func::func_type raft_callback` inside `raft_server::init_options` like this:
```C++
struct init_options {
    // ...

    bool skip_initial_election_timeout_;
    
    cb_func::func_type raft_callback_;
};
```
and invoke `ctx->set_cb_func(...)` is inside the constructor of `raft_server`. Then launcher doesn't need to call `set_cb_func`, thus we can get benefit for both launcher and raft_server. What do you think?

greensky00 (commented):
Thanks @sheepgrass , some minor comments.

include/libnuraft/launcher.hxx
sheepgrass: Agree.

Actually, I was just copied from your unit test case :p
https://github.com/eBay/NuRaft/blob/75017885cd0ad5f10c805f035fa5b44ae991aa3a/tests/unit/raft_package_fake.hxx#L51-L54

greensky00: Oh, good catch, it also needs to be fixed. Thanks :)

src/launcher.cxx
sheepgrass: Sorry that I didn't notice you mentioned about moving it to raft_server constructor.  I will make the change.

Actually, for this change, I am also taken reference to your unit test case:
https://github.com/eBay/NuRaft/blob/75017885cd0ad5f10c805f035fa5b44ae991aa3a/tests/unit/raft_package_fake.hxx#L85-L90

greensky00 (approved):

PR #148, opened 2020-10-20
use find() instead of loop for getting entry from commit_ret_elems_

sheepgrass:
use find() instead of loop for getting entry from commit_ret_elems_


greensky00 (approved):
Thanks @sheepgrass , LGTM. Let me merge it after some internal testing.

Issue #149, opened 2020-10-22
multiple_config_change_test and remove_node_test test cases occasionally failed


sheepgrass:
```
[ .... ] multiple config change test
   === TEST MESSAGE (BEGIN) ===

        time: 2020-10-22 16:36:46.954628
      thread: c858
          in: multiple_config_change_test()
          at: /home/development/Public/NuRaft/tests/unit/raft_server_test.cxx:788
    value of: configs_out.size()
    expected: 3
      actual: 2
[ FAIL ] multiple config change test (728.2 ms)
 [16:36:46.955 825] [tid c858] [FATL] [logger.cc:634, flushAllLoggers()]
Abort
 [16:36:46.956 037] [tid c858] [ERRO] [logger.cc:634, flushAllLoggers()]
 === Critical info (given by user): 0 bytes ===
 [16:36:46.956 167] [tid c858] [ERRO] [logger.cc:634, flushAllLoggers()]
will not explore other threads (disabled by user)
 [16:36:47.196 321] [tid c858] [ERRO] [logger.cc:634, flushAllLoggers()]
```

```
[ .... ] remove node test
   === TEST MESSAGE (BEGIN) ===

        time: 2020-10-22 16:48:42.305966
      thread: 1d15
          in: remove_node_test()
          at: /home/development/Public/NuRaft/tests/unit/raft_server_test.cxx:481
    value of: configs.size()
    expected: 2
      actual: 3
        info: id = 1
[ FAIL ] remove node test (522.2 ms)
 [16:48:42.306 822] [tid 1d15] [FATL] [logger.cc:634, flushAllLoggers()]
Abort
 [16:48:42.307 030] [tid 1d15] [ERRO] [logger.cc:634, flushAllLoggers()]
 === Critical info (given by user): 0 bytes ===
 [16:48:42.307 167] [tid 1d15] [ERRO] [logger.cc:634, flushAllLoggers()]
will not explore other threads (disabled by user)
 [16:48:42.535 961] [tid 1d15] [ERRO] [logger.cc:634, flushAllLoggers()]
```

Seems both cases related to `remove_srv()` and the operation will succeed or fail depending on timing.

greensky00:
Hi @sheepgrass 

This is a known issue due to the asynchronous commit thread. Will provide a deterministic way soon.

In the meantime, increasing `COMMIT_TIME_MS` value to bigger number may help. Thanks.

PR #150, opened 2020-10-23
Make `raft_server_test` and `failure_test` deterministic

greensky00:
* Added a new callback command for state machine execution. Now
`raft_server_test` will not rely on the timer, but deterministically
proceed the test according to the callback signal.

* TODO: Synchronous executor is still based on timing.


greensky00:
#149

PR #152, opened 2020-11-13
[MONSTOR-14106] Correct some potential variable overflow code

shigui1989:


src/handle_snapshot_sync.cxx
```diff
@@ -126,9 +126,9 @@ ptr<req_msg> raft_server::create_sync_snapshot_req(peer& p,
         // LCOV_EXCL_START
         // Raw binary snapshot (original)
         ulong offset = p.get_snapshot_sync_ctx()->get_offset();
-        int32 sz_left = (int32)(snp->size() - offset);
+        ulong sz_left = snp->size() - offset;
```
shigui1989, line 129: @greensky00 I think here we can guarantee that `snp->size() >= offset`?
Or should we use `ulong sz_left = ( ulong sz_left = snp->size() - offset ) ? ( snp->size() - offset ) : 0`?

greensky00: Yes, please go ahead, `sz_left` also should not overflow.

src/handle_join_leave.cxx
```diff
@@ -202,12 +202,14 @@ void raft_server::sync_log_to_new_srv(ulong start_idx) {
     p_db("[SYNC LOG] peer %d start idx %llu, my log start idx %llu\n",
          srv_to_join_->get_id(), start_idx, log_store_->start_index());
     // only sync committed logs
-    int32 gap = (int64_t)quick_commit_index_ - (int64_t)start_idx;
+    ulong gap = ( quick_commit_index_ > start_idx )
+                ? ( quick_commit_index_ - start_idx )
+                : 0;
     ptr<raft_params> params = ctx_->get_params();
     if ( gap < params->log_sync_stop_gap_ ||
```
greensky00, line 209: Now the compiler complains about the comparison between signed/unsigned integers. Let me fix it.

greensky00 (approved):
Thanks @shigui1989

PR #182, opened 2021-02-24
Reset peer's log info once new connection is established

greensky00:
* To support offline data change (including snapshot installation),
peer's log info should be reset if the previous connection is gone.
Otherwise, there will be a data mismatch and the leader will attempt
to send logs based on incorrect info.

* The reset peer info will be recovered during the first
`append_entries` communication between the leader and the peer.

### DIFF
--- a/src/handle_vote.cxx
+++ b/src/handle_vote.cxx
@@ -77,11 +77,7 @@ void raft_server::request_prevote() {
 
             if (recreate) {
                 p_in("reset RPC client for peer %d", s_config->get_id());
-                bool ok = pp->recreate_rpc(s_config, *ctx_);
-                if (ok) {
-                    pp->set_free();
-                    pp->set_manual_free();
-                }
+                pp->recreate_rpc(s_config, *ctx_);
             }
         }
     }
--- a/src/handle_vote.cxx
+++ b/src/handle_vote.cxx
@@ -77,9 +77,11 @@ void raft_server::request_prevote() {
 
             if (recreate) {
                 p_in("reset RPC client for peer %d", s_config->get_id());
-                pp->recreate_rpc(s_config, *ctx_);
-                pp->set_free();
-                pp->set_manual_free();
+                bool ok = pp->recreate_rpc(s_config, *ctx_);
+                if (ok) {
+                    pp->set_free();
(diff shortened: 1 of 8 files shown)


src/handle_vote.cxx
```diff
@@ -77,9 +77,11 @@ void raft_server::request_prevote() {
...
             if (recreate) {
                 p_in("reset RPC client for peer %d", s_config->get_id());
-                pp->recreate_rpc(s_config, *ctx_);
-                pp->set_free();
-                pp->set_manual_free();
+                bool ok = pp->recreate_rpc(s_config, *ctx_);
+                if (ok) {
+                    pp->set_free();
+                    pp->set_manual_free();
+                }
```
sheepgrass, line 84: Seems that every time `recreate_rpc()` call success, `set_free()` and `set_manual_free()` must be called.  Better call them inside `recreate_rpc()` before `return true`.

src/raft_server.cxx
```diff
@@ -893,10 +893,14 @@ void raft_server::reconnect_client(peer& p) {
...
         p_db( "reset RPC client for peer %d",
               p.get_id() );
-        p.recreate_rpc(s_config, *ctx_);
-        p.set_free();
-        p.set_manual_free();
+        bool ok = p.recreate_rpc(s_config, *ctx_);
+        if (ok) {
+            p.set_free();
+            p.set_manual_free();
+        }
```
sheepgrass, line 900: same as previous comment

src/peer.cxx
```diff
@@ -209,10 +209,12 @@ void peer::recreate_rpc(ptr<srv_config>& config,
         //   A reconnection attempt should be treated as an activity,
         //   hence reset timer.
         reset_active_timer();
+        return true;
```
sheepgrass, line 212: better call `set_free()` and `set_manual_free()` here?

sheepgrass (changes requested):
runtests.sh now passed.  Please refer to other comments here.

src/handle_vote.cxx
greensky00: Makes sense, thanks.

yong-li (approved):

PR #187, opened 2021-03-25
Fix rare race condition on commit cv

alesapin:
Usage of condition variables without additional checks almost always leads to bugs. In this case, it may happen that `commit_cv` was notified before `commit_cv_lock_` was acquired. It can lead to a long wait until the next commit notification.


greensky00 (approved):
Thanks, actually it was planned to be replaced with `EventAwaiter` to avoid the situation that you mentioned, and then forgotten. Your PR with predicate is also fine.

PR #191, opened 2021-04-02
Add optional timeout to auto_forwarded requests

alesapin:
Add `auto_forwarding_req_timeout_` setting which cancels forwarded request to the leader after the timeout.

### DIFF
--- a/tests/unit/asio_service_test.cxx
+++ b/tests/unit/asio_service_test.cxx
@@ -31,14 +31,15 @@ namespace asio_service_test {
 
 int launch_servers(const std::vector<RaftAsioPkg*>& pkgs,
                    bool enable_ssl,
-                   bool use_global_asio = false)
+                   bool use_global_asio = false,
+                   const raft_server::init_options & opt = raft_server::init_options())
 {
     size_t num_srvs = pkgs.size();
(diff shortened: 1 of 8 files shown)


tests/unit/asio_service_test.cxx
```diff
@@ -1200,6 +1201,86 @@ int leadership_transfer_test() {
...
+    }
+
+    std::string test_msg = "test";
+    ptr<buffer> msg = buffer::alloc(test_msg.size() + 1);
+    msg->put(test_msg);
+
+    // Forwarded as expected
+    auto ret1 = s3.raftServer->append_entries({msg});
+    CHK_TRUE(ret1->get_accepted());
+    CHK_EQ(ret1->get_result_code(), nuraft::cmd_result_code::OK)
```
greensky00, line 1248: Two parameters need to be switched: `CHK_EQ(expected, variable_to_check)`.
```c++
CHK_EQ(nuraft::cmd_result_code::OK, ret1->get_result_code());
```

greensky00 (approved):
Thanks, looks good to me. A few cosmetic changes are needed, but let me do that in the following commit.

PR #193, opened 2021-04-05
Get rid of auto_destoyer

greensky00:
* The destruction order of global static variables is non-deterministic.

### DIFF
--- a/src/global_mgr.cxx
+++ b/src/global_mgr.cxx
@@ -24,10 +24,41 @@ limitations under the License.
 #include "raft_server.hxx"
 #include "tracer.hxx"
 
+#include <memory>
+
 namespace nuraft {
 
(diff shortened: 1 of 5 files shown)


src/global_mgr.cxx
```diff
@@ -24,10 +24,41 @@ limitations under the License.
...
+    nuraft_global_mgr* get() {
+        return internal_.get();
+    }
+
+    bool create() {
+        if (internal_.get()) {
+            // Already created.
+            return false;
+        }
+        // C++11 doesn't have `make_unique`.
```
alesapin, line 47: Maybe just implement it?)

alesapin (approved):

src/global_mgr.cxx
greensky00: Using constructor with raw pointer is ok for now :)

Issue #200, opened 2021-04-27
Leader election priority test crashes


benstadin:
I've just built NuRaft on OS X 11.3. The election priority test fails as follows:

`[ .... ] leader election priority test
   === TEST MESSAGE (BEGIN) ===

        time: 2021-04-28 01:07:50.099266
      thread: b79a
          in: leader_election_priority_test()
          at: /Users/Ben/dev/NuRaft/tests/unit/raft_server_test.cxx:943
    value of: s2.raftServer->is_leader()
    expected: true
      actual: false
[ FAIL ] leader election priority test (178.1 ms)
 [01:07:50.100 570] [tid b79a] [FATL] [logger.cc:634, flushAllLoggers()]
Abort
 [01:07:50.100 898] [tid b79a] [ERRO] [logger.cc:634, flushAllLoggers()]
 === Critical info (given by user): 0 bytes ===
 [01:07:50.100 970] [tid b79a] [ERRO] [logger.cc:634, flushAllLoggers()]
will not explore other threads (disabled by user)
 [01:07:51.270 971] [tid b79a] [ERRO] [logger.cc:634, flushAllLoggers()]

Thread b79a (crashed here)

#0  0x000000010f7f2c10 in SimpleLoggerMgr::logStackBacktrace(unsigned long) at logger.cc:390
#1  0x000000010f7f32b4 in SimpleLoggerMgr::handleSegAbort(int) at logger.cc:451
#2  0x00007fff2072ad7d in _sigtramp() at libsystem_platform.dylib
#3  0x0000000114416298 in 0x0() at ???
#4  0x00007fff2063a411 in abort() at libsystem_c.dylib
#5  0x000000010f7e7bd0 in TestSuite::reportTestResult(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, int) at test_common.h:1358
#6  0x000000010f7e1572 in main at raft_server_test.cxx:2389
#7  0x00007fff20700f3d in start() at libdyld.dylib
#8  0x0000000000000002 in 0x0() at ???

[ABORT] Flushed all logs safely.
./runtests.sh: line 9:   892 Abort trap: 6           ./tests/raft_server_test --abort-on-failure`

greensky00:
@benstadin 
Thanks for letting us know. This is a test case problem caused by a timing issue. Will fix it soon.

PR #202, opened 2021-05-02
Make fake network transfer order deterministic

greensky00:
* If we use `unordered_map`, its traversal order won't be deterministic
and depending on platforms the test executable is running. We should
make the test deterministic.


greensky00:
https://github.com/eBay/NuRaft/issues/200

PR #213, opened 2021-05-12
Retry snapshot read on the next heartbeat

greensky00:
* If reading snapshot fails, it will be re-attempted upon the next
heartbeat, with the latest (newer) snapshot.


src/handle_snapshot_sync.cxx
```diff
@@ -150,8 +150,18 @@ ptr<req_msg> raft_server::create_sync_snapshot_req(peer& p,
         void*& user_snp_ctx = p.get_snapshot_sync_ctx()->get_user_snp_ctx();
         p_dv("peer: %d, obj_idx: %ld, user_snp_ctx %p\n",
              (int)p.get_id(), obj_idx, user_snp_ctx);
-        state_machine_->read_logical_snp_obj( *snp, user_snp_ctx, obj_idx,
-                                              data, last_request );
+        int rc = state_machine_->read_logical_snp_obj( *snp, user_snp_ctx, obj_idx,
+                                                       data, last_request );
+        if (rc < 0) {
+            p_er( "reading snapshot (idx %lu, term %lu, object %lu) failed: %d",
```
alesapin, line 156: Maybe a warning is enough? It doesn't seem like an unavoidable error, we will retry next time.

src/handle_join_leave.cxx
```diff
@@ -256,10 +256,18 @@ void raft_server::sync_log_to_new_srv(ulong start_idx) {
...
     if (/* start_idx > 0 && */ start_idx < log_store_->start_index()) {
+        srv_to_join_snp_retry_required_ = false;
         req = create_sync_snapshot_req( *srv_to_join_,
                                         start_idx,
                                         state_->get_term(),
                                         quick_commit_index_);
+        if (req == nullptr) {
+            // If reading snapshot fails, enable HB temporarily to retry it.
+            srv_to_join_snp_retry_required_ = true;
+            enable_hb_for_peer(*srv_to_join_);
```
alesapin, line 267: It's not clear to me, but are you sure that these heartbeats cannot lead to some other side effects?

alesapin (approved):

src/handle_join_leave.cxx
greensky00: This is just one-time use with the heartbeat timer. Heartbeat won't be functional (and not be re-scheduled again) as this member is not on the `peers_` list.

src/handle_snapshot_sync.cxx
greensky00: ok.

Issue #241, opened 2021-07-31
The new server cannot find the cluster config when doing snapshot.


nicelulu:
Hi, thank you. We encountered some problems with ```add_srv```. There are now 3 node servers, and it is set to take a snapshot every time 3000000 logs are added. This cluster has been running for a period of time to take a snapshot, and then re-elections were conducted several times, and a new cluster config was generated. Both the config log idx and the previous log idx are greater than 3000000. At this time, a new node is added and the log will be synchronized. When the synchronization reaches 3000000, a snapshot will be taken, but at this time, no cluster config smaller than ```committed_idx``` will be found.

nicelulu:
https://github.com/eBay/NuRaft/pull/242

greensky00:
Hi @nicelulu 
Thanks for bringing this issue.

I'd like to understand what exactly the issue is. Can you please elaborate more?
* If we face the above situation, what is the problem? (crash, system exit, ...)
* Your snapshot distance is `3000000`, what are the values of other parameters? especially `reserved_log_items_` and `log_sync_stop_gap_`.

nicelulu:
Thank you @greensky00.
```reserved_log_items_``` is 10000000
```snapshot_distance``` is 3000000
```log_sync_stop_gap_``` is 99999

At present, the entries in the log_store of the old three-node service are from 0 to 3700000, and the ```log_idx_``` of the current ```cluster_config``` is 3600100, and the ```prev_log_idx_``` is 3600000. After we add a new node, we will first get ```cluster_config``` from the leader, and then synchronize the log from 0. When the log reaches 3000000, the node will do ```snapshot_and_compact```. At this time, it will look for ```cluster_config``` smaller than the current ```committed_idx``` in log_store. Both ```log_idx_``` and ```prev_log_idx_``` of the current ```cluster_config``` of the node exceed the maximum log_idx in ```log_store_```.


https://github.com/eBay/NuRaft/blob/0ffb76b68a2d58174d3c7cd195747b69c2a87ef4/src/handle_commit.cxx#L465

nicelulu:
I have two other questions for advice. 

1. join_cluster_request will synchronize the current cluster_config to the new node. What is the effect of this?
2. What is the practical use of saving last_config_ in snapshot?

greensky00:
@nicelulu 
[...]

PR #244, opened 2021-08-02
Add conf condition at the beginning of snapshot_and_compact.

nicelulu:
Fix https://github.com/eBay/NuRaft/issues/241


greensky00 (approved):
Thanks!

PR #245, opened 2021-08-02
Add lifecycle callback functions for worker threads.

yong-li:
Added lifecycle callback functions for worker-thread start and stop.  The idea behind this PR is the following:
MonstorDB uses user-space update-copy-update (rcu) libraries to manage frequently-read-but-rarely-updated data structures. The library requires each participating thread to call a "register" function before its first use of the library and a "unregister" function before the termination of the thread's main function.  Currently, MonstorDB uses a thread-local boolean flag to track the thread's first use, and relies on another thread-local variable's destructor to call the "unregister" function.  It seems to work at the moment, but theoretically it depends on the order of thread-local variable's destruction order and may break in future in a very subtle way.  I'm planning to fix that by explicitly calling "register" and "unregister" functions inside each thread's main function (for threads using the rcu library only).  Currently, the "pre_commit" callback functions may use the rcu library.  Therefore, I added the two hooks on the worker thread's main function.

### DIFF
--- a/src/asio_service.cxx
+++ b/src/asio_service.cxx
@@ -1590,12 +1590,17 @@ std::string asio_service_impl::get_password
 #endif
 
 void asio_service_impl::worker_entry() {
-    std::string thread_name = "nuraft_w_" + std::to_string(worker_id_.fetch_add(1));
+    uint32_t worker_id = worker_id_.fetch_add(1);
+    std::string thread_name = "nuraft_w_" + std::to_string(worker_id);
 #ifdef __linux__
     pthread_setname_np(pthread_self(), thread_name.c_str());
 #elif __APPLE__
     pthread_setname_np(thread_name.c_str());
 #endif
+
+    if (my_opt_.worker_start_) {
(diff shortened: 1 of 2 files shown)


adamsofea1111 (approved):

greensky00 (approved):

src/asio_service.cxx
```diff
@@ -1590,13 +1590,18 @@ std::string asio_service_impl::get_password
...
+    uint32_t worker_id = worker_id_.fetch_add(1);
+    std::string thread_name = "nuraft_w_" + std::to_string(worker_id);
 #ifdef __linux__
     pthread_setname_np(pthread_self(), thread_name.c_str());
 #elif __APPLE__
     pthread_setname_np(thread_name.c_str());
 #endif
 
+    if (my_opt_.worker_start_) {
+        my_opt_.worker_start_(worker_id);
```
greensky00, line 1602: Not sure `worker_id` is what you intended (it is different from thread ID). Let me merge it for now.

greensky00 (approved):

PR #278, opened 2022-01-22
Fix bug in `commit_ret_elems` which lead to "live leak" in case of requests batch.

### LINKED ISSUE #239: A lot of "cancelled non-blocking client request" with async API
It seems like during server start when we apply non-snapshotted log records to the state machine we add elements to the `commit_ret_elems_` https://github.com/eBay/NuRaft/blame/master/src/handle_commit.cxx#L353 even for async_handler API. But nobody will erase them, and during the shutdown, we will get a lot of `cancelled non-blocking client request` in `drop_all_pending_commit_elems`.

BTW I don't understand why we care about `commit_ret_elems_` for non-blocking mode: https://github.com/ebay/NuRaft/blob/master/include/libnuraft/raft_server.hxx#L1270

alesapin:
I think the code contains explicit comment. Fixes #239.

### DIFF
--- a/src/handle_commit.cxx
+++ b/src/handle_commit.cxx
@@ -300,14 +300,27 @@ void raft_server::commit_app_log(ulong idx_to_commit,
     std::list< ptr<commit_ret_elem> > async_elems;
     if (need_to_handle_commit_elem) {
         std::unique_lock<std::mutex> cre_lock(commit_ret_elems_lock_);
-        bool match_found = false;
+        /// Sometimes user can batch requests to RAFT: for example send 30
+        /// append entries requests in a single batch. For such request batch
+        /// user will receive a single response: all was successful or all
+        /// failed. Obviously we don't need to add info about responses
+        /// (commit_ret_elems) for 29 requests from batch and need to do it only
+        /// for 30-th request. precommit_index is exact value which identify ID
+        /// of the last request from the latest batch. So if we commiting this
+        /// last request and for some reason it was not added into
+        /// commit_ret_elems in the handle_cli_req method (logical race
+        /// condition) we have to add it here. Otherwise we don't need to add
+        /// anything into commit_ret_elems_, because nobody will wait for the
+        /// responses of the intermediate requests from requests batch.
+        bool need_to_check_commit_ret = sm_idx == pc_idx;
+
         auto entry = commit_ret_elems_.find(sm_idx);
         if (entry != commit_ret_elems_.end()) {
             ptr<commit_ret_elem> elem = entry->second;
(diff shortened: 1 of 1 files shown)


src/handle_commit.cxx
```diff
@@ -300,14 +300,27 @@ void raft_server::commit_app_log(ulong idx_to_commit,
...
+        /// failed. Obviously we don't need to add info about responses
+        /// (commit_ret_elems) for 29 requests from batch and need to do it only
+        /// for 30-th request. precommit_index is exact value which identify ID
+        /// of the last request from the latest batch. So if we commiting this
+        /// last request and for some reason it was not added into
+        /// commit_ret_elems in the handle_cli_req method (logical race
+        /// condition) we have to add it here. Otherwise we don't need to add
+        /// anything into commit_ret_elems_, because nobody will wait for the
+        /// responses of the intermediate requests from requests batch.
+        bool need_to_check_commit_ret = sm_idx == pc_idx;
```
greensky00, line 315: Thanks for this fix. To make sure that I understand this condition correctly, can you confirm the below?

Let's say the current committed index is 10 (for all pre-commit, quick-commit, and sm-commit), and two threads (T1 and T2) are calling `handle_cli_req` with exactly one log. What will happen is:
1. T1 calls `handle_cli_req` first, `precommit_index_` becomes 11.
2. Before T1 creates the corresponding `commit_ret_elem`, the commit thread wakes up earlier.
3. Since `sm_idx == pc_idx` in this case, the commit thread will create it.
4. However, T2 cannot interfere before step (3), because of the lock held by `handle_cli_req_prelock`. T2 can append its log only after the `commit_ret_elem` for 11 is created.

So, if a thread order inversion happens, only one user thread is involved at a time. When the commit thread wakes up, there is no case that there are multiple `handle_cli_req` calls whose `commit_ret_elem`s are not created yet. Is that right?

greensky00 (approved):

src/handle_commit.cxx
alesapin: Actually I didn't think about such case, but yes `handle_cli_req` guarded by mutex (or recursive mutex for some reason...) in `handle_cli_req_prelock`:  https://github.com/ClickHouse-Extras/NuRaft/blob/1707a7572aa66ec5d0a2dbe2bf5effa3352e6b2d/src/handle_client_request.cxx#L40-L53.

alesapin: BTW I cannot catch "thread inversion" even under heavy load. But according to the code it should be possible.

greensky00: Thanks, let me merge it.

PR #290, opened 2022-02-24
Fix log output

JackyWoo:
Found an overflow int32 logging.

```
2 nodes (out of 3, 3 including learners) are not responding longer than 10000 ms, at least 1645066963648514 node        s (including leader) should be alive to proceed commit
```


JackyWoo:
Maybe it is not a good way or logging because too easy to miss match type.

greensky00:
Can you fix other formats in the same log as well?
```c++
        p_er("%d nodes (out of %d, %zu including learners) are not "
             "responding longer than %d ms, "
             "at least %d nodes (including leader) should be alive "
             "to proceed commit",
```

greensky00 (approved):
LGTM

PR #294, opened 2022-03-01
Fix to race between `update_term` and `handle_cli_req`

greensky00:
* Server role should be changed BEFORE accepting the newer term.


JackyWoo:
LGTM

Issue #298, opened 2022-03-03
Client thread return error if gap between log store and state machine is too large in leader.


JackyWoo:
When log replication is slow the gap will be large whih may lead to client timeout. As my understanding when timeout occur client does not know what is happening whether the request will be commited or lose.  So we'd better avoid timeout and return unambiguous response to client.

Maybe we can add a limit to gap between log store and state machine, when reaches return error code to client.

Correct me if my understanding is wrong.

If the idea is ok I'd like to work on it.

greensky00:
Similar to https://github.com/eBay/NuRaft/issues/297, that should be done by the application (the user of NuRaft). The application is able to know the gap by APIs provided by NuRaft and can fail fast without calling the `append_entries()` API.

JackyWoo:
Got it  thanks

Issue #297, opened 2022-03-03
Limit client append entry size


JackyWoo:
I wonder whether we should limit the client append entry size. If not there will be extremely large entries theoretically, which will lead a client log batch will be distributed to followers by more than one batch. Maybe this will lead to something bad.

I notice there is a config `max_append_size_` who will limit log size to distribute to follower once, but it does limit the client appending size. 

1. So maybe we can use it or
2. Add a new one  `max_client_append_size_` for example.

greensky00:
That (having config for the max appending logs) should be done by the application (the user of NuRaft) before calling `append_entries()` API, not by NuRaft.

PR #299, opened 2022-03-03
Add log to append entries API

JackyWoo:
Add some log to append entries API to help users to know how to use it more correctly. Bellow two issue is related #298 #297

### DIFF
--- a/include/libnuraft/raft_server.hxx
+++ b/include/libnuraft/raft_server.hxx
@@ -220,6 +220,11 @@ public:
      * Append and replicate the given logs.
      * Only leader will accept this operation.
      *
+     * Users need to take care of the 2 points
+     *    1. Limit the logs size in one batch
+     *    2. Limit the speed of appending entries,
+     *       user can get the gap of log store and state machine by NuRaft API.
+     *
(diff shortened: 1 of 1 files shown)


include/libnuraft/raft_server.hxx
```diff
@@ -220,6 +220,11 @@ public:
      * Append and replicate the given logs.
      * Only leader will accept this operation.
      *
+     * Users need to take care of the 2 points
+     *    1. Limit the logs size in one batch
+     *    2. Limit the speed of appending entries,
+     *       user can get the gap of log store and state machine by NuRaft API.
```
greensky00, line 226: I don't think this additional statement is necessary, as it is too trivial and use-case-specific. 

For example, think about `pwrite` API, it sounds like
* Users should limit the size of buffer passed to `pwrite`, or it may cause IO failure.
* Users should limit the speed of disk writes, or it may cause IO failure.

"How to handle failure" is totally up to users. Someone will be ok with the above cases with their special log store and state machine implementations.

JackyWoo: ok

Issue #347, opened 2022-04-12
When using Auto forwarding and async_handler, result is not available after calling append_entries from follower.


nicelulu:
async_handler：

1. calling append_entries from follower:
```
nodes：A, B, C. 

B is leader

A -> B send client request, then do something ..., result.get() /// wait

B process_req() /// block mode wait commit then send response, async mode send response, no result yet.

when B -> A resp, A presult->set_result(resp_ctx, perr) /// notify
```

2. calling append_entries from leader：
```
nodes：A, B, C. 

B is leader

client -> B client request, then do something ..., result.get() /// wait

B process_req() /// block mode wait commit then return result, async mode return result

B handle_commit. elem->async_result_->set_result( elem->ret_value_, err, cmd_result_code::OK ) /// notify
```

If the follower uses the async_handler mode, here the leader needs to block until it gets the result?

greensky00:
I understand the problem that you described, but your PR is not a proper way as it blocks Asio worker threads. If you don't mind, let me fix this problem as it seems to require restructuring the Asio service layer.

nicelulu:
@greensky00 Thanks.

greensky00:
https://github.com/eBay/NuRaft/pull/350

PR #350, opened 2022-04-14
Deliver commit results for auto-forwarding in async mode

greensky00:
* In async mode, the commit result will be ready later. If the request
is come from the other server through auto-forwarding, the request
should wait for the result before returning the response.


greensky00:
https://github.com/eBay/NuRaft/issues/347

PR #358, opened 2022-06-02
Use find_package for OpenSSL

ppande:
* Remove hardcoded paths and use find_package


greensky00:
Before diving into the review, what is the purpose of this commit? Building NuRaft has no problem with openssl, and if you want to define a custom path, you can simply do 
```
cmake -DDEPS_PREFIX=<path>
```

And FYI, this PR causes an error on Mac as follows, even though openssl is already installed in `/usr/local/opt/openssl`:
```
-- deps prefix is not given
CMake Error at /usr/local/Cellar/cmake/3.15.4/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find OpenSSL, try to set the path to OpenSSL root folder in the
  system variable OPENSSL_ROOT_DIR (missing: OPENSSL_CRYPTO_LIBRARY
  OPENSSL_INCLUDE_DIR)
Call Stack (most recent call first):
  /usr/local/Cellar/cmake/3.15.4/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:378 (_FPHSA_FAILURE_MESSAGE)
  /usr/local/Cellar/cmake/3.15.4/share/cmake/Modules/FindOpenSSL.cmake:413 (find_package_handle_standard_args)
  CMakeLists.txt:89 (find_package)


-- Configuring incomplete, errors occurred!
```

ppande:
>  Before diving into the review, what is the purpose of this commit? Building NuRaft has no problem with openssl, and if you want to define a custom path, you can simply do
The purpose is to replace any hardcoded paths or custom variables with built-in support provided by [cmake for OpenSSL] (https://cmake.org/cmake/help/latest/module/FindOpenSSL.html), and in this case specifically for OpenSSL. The problem arises when it is one of the various dependencies in a project.

>  And FYI, this PR causes an error on Mac as follows, ..
As the message reported by cmake describes, setting `OPENSSL_ROOT_DIR` to `/usr/local/opt/openssl` would get you past it

greensky00:
There are two problems that must be fixed:

* Custom path should work to choose the preferred SSL library:
  * If someone does `cmake -DDEPS_PREFIX=<path> ..`, then the library in `<path>` should be chosen rather than the system-installed one.
* Clean build from scratch should work without any custom variables.
  * `brew install openssl` and then `cmake ..` should work without passing `-DOPENSSL_ROOT_DIR=<path>`.

Once those are fixed, let me start the review of the CMake script itself. Thanks.

greensky00:
* Still `-DDEPS_PREFIX=` is ignored on Linux, your new change takes effect on `APPLE` only.
* Still `cmake` without `OPENSSL_ROOT_DIR` does not work on Mac. Looks like it is a [known issue of CMake find_package for Homebrew-installed OpenSSL](https://www.scivision.dev/cmake-macos-find-openssl/). Without fixing it, it will break existing builds and force them to add `-DOPENSSL_ROOT_DIR=` which will not be acceptable. 
* If OpenSSL is not found in the given path, the current NuRaft just moves forward and uses `-lssl -lcrypto` flags to link system libraries. But with your change, it immediately fails at `find_package` command. It will also not be acceptable.

ppande:
Thanks @greensky00. I will re-work these changes to incorporate your suggestions and update the PR

JosiahWI:
Thank you for starting these changes, @ppande. There wasn't any progress on this for over a year, so I picked up the work and #451 has now landed. I think this PR can be closed.

PR #363, opened 2022-06-25
fix segfault on 32bit system: fix #243, fix #362

### LINKED ISSUE #243: compiling nuraft for the raspberry pi
I am trying to compile nuraft for the raspberry pi
`root ~/NuRaft/build >g++ -v
Using built-in specs.
COLLECT_GCC=g++
COLLECT_LTO_WRAPPER=/usr/lib/gcc/arm-linux-gnueabihf/8/lto-wrapper
Target: arm-linux-gnueabihf
Configured with: ../src/configure -v --with-pkgversion='Raspbian 8.3.0-6+rpi1' --with-bugurl=file:///usr/share/doc/gcc-8/README.Bugs --enable-languages=c,ada,c++,go,d,fortran,objc,obj-c++ --prefix=/usr --with-gcc-major-version-only --program-suffix=-8 --program-prefix=arm-linux-gnueabihf- --enable-shared --enable-linker-build-id --libexecdir=/usr/lib --without-included-gettext --enable-threads=posix --libdir=/usr/lib --enable-nls --enable-bootstrap --enable-clocale=gnu --enable-libstdcxx-debug --enable-libstdcxx-time=yes --with-default-libstdcxx-abi=new --enable-gnu-unique-object --disable-libitm --disable-libquadmath --disable-libquadmath-support --enable-plugin --with-system-zlib --with-target-system-zlib --enable-objc-gc=auto --enable-multiarch --disable-sjlj-exceptions --with-arch=armv6 --with-fpu=vfp --with-float=hard --disable-werror --enable-checking=release --build=arm-linux-gnueabihf --host=arm-linux-gnueabihf --target=arm-linux-gnueabihf
Thread model: posix
gcc version 8.3.0 (Raspbian 8.3.0-6+rpi1) 
`
Please find attached the build log.
The build fails when I compile on the raspberry or during cross compiling
[build.log](https://github.com/eBay/NuRaft/files/6910845/build.log)

### LINKED ISSUE #362: segfault when cross-compiled to ARM device
Sorry to bother. I want to use NuRaft in a embedded device because NuRaft is so small and lightweight compared to braft, and it has only one external dependencies which makes it perfect for cross compilation. My device requires me to use a special cross compiler called `arm-poky-linux-gnueabi-gcc` to compile it, and I write a whole set of CMake TOOLCHAIN.cmake to configure everything I need, including OpenSSL lib path, etc. Everything compiled, and `make` exited with `0` returned. But when I try to execute the compiled tests on the device, **it segfault immediately**:

image

test `buffer_test`, `serialization_test`, `timer_test`, `strfmt_test` and `stat_mgr_test` works fine and all passed, but `raft_server_test`, `failure_test` and `asio_service_test`, all failed with segfault.

Interestingly the stacktrace they printed are looked pretty similar. Here's the `addr2line`-ed version of the dump file of `./tests/asio_service_test`:
```
root@ATK-IMX6U:~/NuRaft# cat dump_20210222_230807+0800.txt 
When: 2021-02-22T23:08:07.549757+08:00

 === Critical info (given by user): 0 bytes ===

will not explore other threads (disabled by user)

Thread fd71 1141
(crashed here)
0x8af00 (/home/ray-eldath/CodeProjects/lajishixun2/NuRaft/examples/logger.h:343)
0x8ca0c (/home/ray-eldath/CodeProjects/lajishixun2/NuRaft/examples/logger.cc:393)
0x8cebc (/home/ray-eldath/CodeProjects/lajishixun2/NuRaft/examples/logger.cc:431)
0x76a6a880 (??:0)


Thread fd71 (1141) (crashed here)

#0  0x000000000008ca0c in L�t���v at x0r8P�t
#1  0x000000000008cebc in L�t���v at x0r8P�t
#2  0x0000000076a6a880 in L�t���v at x0r8P�t
```
And here's another dump file of `./tests/raft_server_test`:
```
root@ATK-IMX6U:~/NuRaft# cat dump_20210222_230008+0800.txt 
When: 2021-02-22T23:00:08.993242+08:00

 === Critical info (given by user): 0 bytes ===

will not explore other threads (disabled by user)

Thread c77c 1109
(crashed here)
0x95f00 (/home/ray-eldath/CodeProjects/lajishixun2/NuRaft/examples/logger.h:343)
0x97a0c (/home/ray-eldath/CodeProjects/lajishixun2/NuRaft/examples/logger.cc:393)
0x97ebc (/home/ray-eldath/CodeProjects/lajishixun2/NuRaft/examples/logger.cc:431)
0x76c3c880 (??:0)


Thread c77c (1109) (crashed here)

#0  0x0000000000097a0c in �t�~���v() at `���x�~
#1  0x0000000000097ebc in _ZN15SimpleLoggerMgr14handleSegFaultEi() at `���x�~
#2  0x0000000076c3c880 in __default_sa_restorer() at `���x�~
```

It seems that all these segfaults are triggered by the same bug in `examples/logger.h` and `examples/logger.cc`.

On my local computer (x86) everything works fine. But when cross-compiled everything just segfault-ed. Are there any fix to such segfault? many thx.

Ray-Eldath:
this PR fixes the unportability of NuRaft, now NuRaft could finally works on 32bit systems.

1. core problem is the widely-used typedef `nuraft::ulong` is defined as `uint64_t`, which is 8 bytes. on 64bit systems, 8B integer is defined as `unsigned long` which could be printed with `%ld` / `%lu`, but on 32bit systems, 8B integer should only be printed with `%lld` / `%llu`, if not doing so, the following data would have wrong offset, **causing the whole program be terminated by segfault.**

2. another problem is that NuRaft use a lot of `%zu` / `%z` (which is print a `size_t`) to print `nuraft:ulong`, **this could also cause segfault on 32bit systems**, since `size_t` is actually _shorter than_ 8B, as demonstrated in the following demo:

on armhf SoC and x86_32 (32bit):

```
root@ATK-IMX6U:~# ./test
long int=4  uint64_t=ulong=8
long long int=8  uint64_t=ulong=8
size_t=4  uint64_t=ulong=8
```

on aarch64 and x86_64 (64bit):

```
<~/C/l>-> ./test.x86
long int=8  uint64_t=ulong=8
long long int=8  uint64_t=ulong=8
size_t=8  uint64_t=ulong=8
```

---

this PR fix both bug (1) and (2):
1. for bug (1), there're several possible solutions:
   - the most simple solution is to just change `nuraft::ulong` to platform-dependent `unsigned long`, thus fits the `%ld` used. but this has an obvious adverse effect on Raft protocol, since 32bit is very prone to overflow, especially in some mission-critical storage system that may under heavy load just as NuRaft.
   - [`inttypes.h`](https://en.cppreference.com/w/c/types/integer) have several platform-dependent macros we can use to properly print a 64bit integer on both 32bit and 64bit platforms. we can also automatically replace `%ld` / `%lu` to `"%" PRId64` / `"%" PRId64`. this could be done at compile-time or run-time. unfortunately the vararg macros of NuRaft's logger seems incapable of processing C's string concat syntax (`"%" "d" == "%d"`), thus this PR implement this fix at run-time, which leads to a very small performance degrade on 32bit systems. considering 32bit systems're really rare these days and people usually don't use a very noisy log level, I think this solution's OK.
   - we can also do such global replace (`%ld -> %lld`, `%lu -> %llu`) in _build system_. but I'm not very familiar with CMake, and comparing to the two solutions stated above which fix the bug _by code_, I dont think fix _by build system_ (something that is _not_ code) is a great idea.
   - finally, we can document such global replace in README.md to those rare 32bit platform users that they need to do two global replacement _before_ building NuRaft.
2. for bug (2), this PR correct some misuse of `%zu` and `%z`. this is **necessary** and must be settle down, and from now on NuRaft mustn't introduce new code that misuses `%zu` or `%z`. I havn't check every `%zu` and `%z`, I just fix some misuses of them that would cause the tests to segfault, this means maybe there're many of them still lurking in the code.

---

more details could be found in #362.

### DIFF
--- a/tests/test_common.h
+++ b/tests/test_common.h
@@ -1009,9 +1009,10 @@ public:
                 std::string _unit =
                     (unit.empty()) ? "" : unit + " ";
 
-                _msg("\r%s%ld/%ld %s(%.1f%%)",
-                     _comment.c_str(), curValue, num, _unit.c_str(),
-                     (double)curValue*100/num);
+                Msg mm;
+                mm << "\r" << _comment << curValue << "/" << num << " " << _unit
+                   << std::fixed << std::setprecision(1) << "("
+                   << (double)curValue * 100 / num << "%)";
                 fflush(stdout);
             }
             if (curValue >= num) {
--- a/tests/unit/asio_service_test.cxx
+++ b/tests/unit/asio_service_test.cxx
@@ -107,10 +107,10 @@ int make_group_test() {
         uint64_t last_log_idx = s1.raftServer->get_last_log_idx();
         CHK_EQ(srv_id, pi.id_);
         CHK_EQ(last_log_idx, pi.last_log_idx_);
-        _msg("srv %d: %lu, responeded %.1f ms ago\n",
-             pi.id_,
-             pi.last_log_idx_,
-             pi.last_succ_resp_us_ / 1000.0);
+        TestSuite::Msg mm;
+        mm << "srv " << pi.id_ << ": " << pi.last_log_idx_ << ", responded " << std::fixed
+           << std::setprecision(1) << pi.last_succ_resp_us_ / 1000.0 << " ms ago"
+           << std::endl;
     }
 
     // Sleep a while and get all info.
@@ -121,10 +121,10 @@ int make_group_test() {
     for (raft_server::peer_info& pi: v_pi) {
         uint64_t last_log_idx = s1.raftServer->get_last_log_idx();
         CHK_EQ(last_log_idx, pi.last_log_idx_);
-        _msg("srv %d: %lu, responeded %.1f ms ago\n",
-             pi.id_,
-             pi.last_log_idx_,
-             pi.last_succ_resp_us_ / 1000.0);
+        TestSuite::Msg mm;
+        mm << "srv " << pi.id_ << ": " << pi.last_log_idx_ << ", responded " << std::fixed
+           << std::setprecision(1) << pi.last_succ_resp_us_ / 1000.0 << " ms ago"
+           << std::endl;
     }
 
     s1.raftServer->shutdown();
@@ -2653,14 +2653,15 @@ int main(int argc, char** argv) {
 #else
     _msg("raft stats: DISABLED\n");
 #endif
-    _msg("num allocs: %zu\n"
-         "amount of allocs: %zu bytes\n"
-         "num active buffers: %zu\n"
-         "amount of active buffers: %zu bytes\n",
-         raft_server::get_stat_counter("num_buffer_allocs"),
-         raft_server::get_stat_counter("amount_buffer_allocs"),
-         raft_server::get_stat_counter("num_active_buffers"),
-         raft_server::get_stat_counter("amount_active_buffers"));
+    TestSuite::Msg mm;
+    mm << "num allocs: " << raft_server::get_stat_counter("num_buffer_allocs")
+       << std::endl
+       << "amount of allocs: " << raft_server::get_stat_counter("amount_buffer_allocs")
+       << " bytes" << std::endl
+       << "num active buffers: " << raft_server::get_stat_counter("num_active_buffers")
+       << std::endl
+       << "amount of active buffers: "
+       << raft_server::get_stat_counter("amount_active_buffers") << " bytes" << std::endl;
 
     return 0;
 }
--- a/src/asio_service.cxx
+++ b/src/asio_service.cxx
@@ -245,7 +245,7 @@ public:
 
         cached_address_ = socket_.remote_endpoint().address().to_string();
         cached_port_ = socket_.remote_endpoint().port();
-        p_in( "session %zu got connection from %s:%u (as a server)",
+        p_in( "session %lu got connection from %s:%u (as a server)",
               session_id_,
               cached_address_.c_str(),
               cached_port_ );
@@ -269,14 +269,14 @@ public:
     void handle_handshake(ptr<rpc_session> self,
                           const ERROR_CODE& err) {
         if (!err) {
-            p_in( "session %zu handshake with %s:%u succeeded (as a server)",
+            p_in( "session %lu handshake with %s:%u succeeded (as a server)",
                   session_id_,
                   cached_address_.c_str(),
                   cached_port_ );
             this->start(self);
 
         } else {
-            p_er( "session %zu handshake with %s:%u failed: error %d, %s",
+            p_er( "session %lu handshake with %s:%u failed: error %d, %s",
                   session_id_,
                   cached_address_.c_str(),
                   cached_port_,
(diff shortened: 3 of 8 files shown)


tests/test_common.h
```diff
@@ -825,6 +827,7 @@ class TestSuite {
                (cur_test->options.printTestMessage || cur_test->displayMsg) &&
                !cur_test->suppressMsg ) ||
              globalMsgFlag() ) {
+            _fix_format(format);
```
greensky00, line 830: Unlike the logger in NuRaft, both `_msg` and `_msgt` are just wrappers around traditional `printf` so it should behave exactly the same as `printf` without `_fix_format`. Instead, please refer my comment below.

tests/unit/asio_service_test.cxx
```diff
@@ -107,7 +107,7 @@ int make_group_test() {
         uint64_t last_log_idx = s1.raftServer->get_last_log_idx();
         CHK_EQ(srv_id, pi.id_);
         CHK_EQ(last_log_idx, pi.last_log_idx_);
-        _msg("srv %d: %lu, responeded %.1f ms ago\n",
+        _msg("srv %d: %llu, responeded %.1f ms ago\n",
```
greensky00, line 110:
```c++
TestSuite::Msg mm;
mm << "srv " << pi.id_ << ": " << pi.last_log_idx_ << ", responsded "
   << std::fixed << std::setprecision(1) << pi.last_succ_resp_us_ / 1000.0
   << " ms ago" << std::endl;
```

include/libnuraft/fix_format.hxx
```diff
@@ -0,0 +1,25 @@
+#pragma once
```
greensky00, line 1: Please move `fix_format.hxx` file to `/src`, as it is not for public purpose.

include/libnuraft/fix_format.hxx
```diff
@@ -0,0 +1,25 @@
...
+    }
+}
+
+#define _fix_format(format)                        \
+    std::string format_string(format);             \
+    replace_all(format_string, "%ld", "%" PRId64); \
+    replace_all(format_string, "%lu", "%" PRIu64); \
+    format = format_string.c_str();
+#else
+#define _fix_format(format) do {} while(0)
```
greensky00, line 24: I think simply putting nothing will be fine:
```c++
#define _fix_format(format)
```

tests/unit/asio_service_test.cxx
```diff
@@ -121,7 +121,7 @@ int make_group_test() {
     for (raft_server::peer_info& pi: v_pi) {
         uint64_t last_log_idx = s1.raftServer->get_last_log_idx();
         CHK_EQ(last_log_idx, pi.last_log_idx_);
-        _msg("srv %d: %lu, responeded %.1f ms ago\n",
+        _msg("srv %d: %llu, responeded %.1f ms ago\n",
```
greensky00, line 124: Ditto.

greensky00 (commented):
Thanks for your contribution @Ray-Eldath. Overall looks good to me, and I put some minor comments.

Ray-Eldath:
just fixed what you've said, and still passed all tests on my 32bit arm device and 64bit x86_64. would you mind take a look again? @greensky00

greensky00 (approved):
LGTM, thanks!

PR #366, opened 2022-07-28
Initialize parallel_log_appending

antonio2368:
Really small fix to make it more consistent and to make MSAN happy 😄 (it was caught in our CI https://github.com/ClickHouse/ClickHouse/pull/39609)


greensky00 (approved):
Thanks for catching this!

PR #370, opened 2022-08-22
Windows does not include POSIX gettimeofday() call.

szmyd:
Convert call to gettimeofday() (POSIX only) to use C++11 APIs available on all platforms.

### DIFF
--- a/include/libnuraft/internal_timer.hxx
+++ b/include/libnuraft/internal_timer.hxx
@@ -121,11 +121,9 @@ struct timer_helper {
 
     static uint64_t get_timeofday_us() {
         namespace sc = std::chrono;
-        sc::system_clock::duration d = sc::system_clock::now().time_since_epoch();
-        sc::seconds s = sc::duration_cast<sc::seconds>(d);
-        uint64_t ret = s.count() * 1000000UL;
-        ret += sc::duration_cast<sc::microseconds>(d - s).count();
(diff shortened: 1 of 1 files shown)


include/libnuraft/internal_timer.hxx
```diff
@@ -122,10 +120,11 @@ struct timer_helper {
...
     static uint64_t get_timeofday_us() {
-        struct timeval tv;
-        gettimeofday(&tv, nullptr);
-        uint64_t ret = tv.tv_sec * 1000000UL;
-        ret += tv.tv_usec;
+        namespace sc = std::chrono;
+        sc::system_clock::duration d = sc::system_clock::now().time_since_epoch();
+        sc::seconds s = sc::duration_cast<sc::seconds>(d);
+        uint64_t ret = s.count() * 1000000UL;
+        ret += sc::duration_cast<sc::microseconds>(d - s).count();
```
greensky00, line 127: Can we just do it as follows?
```c++
sc::microseconds s = sc::duration_cast<sc::microseconds>(d);
uint64_t ret = s.count();
```

szmyd: Yes, this Copy/Paste from other code that kept them separate. Will change.

greensky00 (approved):
Thanks!

PR #372, opened 2022-08-26
avoid simultaneously snapshot

### LINKED ISSUE #371: Avoid simultaneously snapshot.
### Issue
Avoid simultaneously snapshot to avoid all nodes being not available in a moment for  long time snapshoting.

NuRaft provides `chk_create_snapshot` method with which user can determine when to snapshot. But maybe it is more user friendly to make it default.  

So @greensky00 what is your opinion?

### Solution

Add `cur_snapshot_distance_` to `raft_server` and set it to `params->snapshot_distance_ /2 + random(params->snapshot_distance_ / 2)`

JackyWoo:
This PR try to close #371

### DIFF
--- a/src/handle_commit.cxx
+++ b/src/handle_commit.cxx
@@ -468,19 +468,18 @@ bool raft_server::snapshot_and_compact(ulong committed_idx, bool forced_creation
         return false;
     }
 
+    auto snapshot_distance = (ulong)params->snapshot_distance_;
+    // Randomized snapshot distance for the first creation.
+    if ( params->enable_randomized_snapshot_creation_ &&
+        !snp_in_progress_.load(std::memory_order_relaxed) &&
+        !get_last_snapshot() &&
+        params->snapshot_distance_ != 0 ) {
+
+        snapshot_distance = first_snapshot_distance_;
+    }
+
     if (!forced_creation) {
         // If `forced_creation == true`, ignore below conditions.
-        auto snapshot_distance = (ulong)params->snapshot_distance_;
-
-        // Randomized snapshot distance for the first creation.
-        if ( params->enable_randomized_snapshot_creation_ &&
-            !snp_in_progress_.load(std::memory_order_relaxed) &&
-            !get_last_snapshot() &&
-            params->snapshot_distance_ != 0 ) {
-
-            snapshot_distance = first_snapshot_distance_;
-        }
-
         if ( params->snapshot_distance_ == 0 ||
              ( committed_idx - log_store_->start_index() + 1 ) < snapshot_distance ) {
             // snapshot is disabled or the log store is not long enough
@@ -499,8 +498,7 @@ bool raft_server::snapshot_and_compact(ulong committed_idx, bool forced_creation
     ptr<snapshot> local_snp = get_last_snapshot();
     if ( ( forced_creation ||
            !local_snp ||
-           ( committed_idx - local_snp->get_last_log_idx() ) >=
-                 (ulong)params->snapshot_distance_ ) &&
+           ( committed_idx - local_snp->get_last_log_idx() ) >= snapshot_distance ) &&
          snp_in_progress_.compare_exchange_strong(f, true) )
     {
         snapshot_in_action = true;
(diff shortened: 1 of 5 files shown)


JackyWoo:
@greensky00 I wonder whether we should add testing to CI.

src/handle_commit.cxx
```diff
@@ -469,12 +470,30 @@ bool raft_server::snapshot_and_compact(ulong committed_idx, bool forced_creation
...
+            uint seed = (uint)( std::chrono::system_clock::now()
+                                   .time_since_epoch().count() * id_ );
+            std::default_random_engine engine(seed);
+            std::uniform_int_distribution<int32>
+                distribution( 0, params->snapshot_distance_ / 2 );
+
+            snapshot_distance = params->snapshot_distance_ / 2
+                                + distribution(engine);
+
+            p_in("First snapshot creation log distance %llu", snapshot_distance);
```
greensky00, line 488: It will not provide enough randomness, as mostly all members will hit this logic at the same time.

I recommend defining a member variable
```c++
    /**
     * (Read-only)
     * The first snapshot distance.
     */
    uint32_t first_snapshot_distance_;
```
and initialize it in the constructor of `raft_server` like this, based on the timestamp of Raft server initialization:
```c++
, first_snapshot_distance_(ctx_->get_params()->snapshot_distance_)
...

    if (params->stale_log_gap_ < params->fresh_log_gap_) {
        params->stale_log_gap_ = params->fresh_log_gap_;
    }
    if (params->enable_randomized_snapshot_creation_) {
        uint64_t seed = timer_helper::get_timeofday_us() * id_;

        // Flip the integer.
        uint8_t* first = reinterpret_cast<uint8_t*>(&seed);
        uint8_t* last = reinterpret_cast<uint8_t*>(&seed) + sizeof(seed);
        while ((first != last) && (first != -- last)) std::swap(*first++, *last);

        std::default_random_engine engine(seed);
        std::uniform_int_distribution<int32>
            distribution( params->snapshot_distance_ / 2, params->snapshot_distance_ );

        first_snapshot_distance_ = distribution(engine);
        p_in("First snapshot creation log distance %llu", first_snapshot_distance_);
    }
```

And this if-clause just does
```c++
snapshot_distance = first_snapshot_distance_;
```

JackyWoo:
@greensky00 Pls review if you have a chance.

tests/unit/raft_server_test.cxx
```diff
@@ -2027,6 +2027,92 @@ int snapshot_manual_creation_test() {
...
+    s1.fNet->execReqResp(); // commit.
+    CHK_Z( wait_for_sm_exec(pkgs, COMMIT_TIMEOUT_SEC) ); // commit execution.
+
+    // Remember the current commit index.
+
+    uint64_t snp_last_log_idx1 = s1.getTestSm()->last_snapshot()->get_last_log_idx();
+    uint64_t snp_last_log_idx2 = s2.getTestSm()->last_snapshot()->get_last_log_idx();
+    uint64_t snp_last_log_idx3 = s3.getTestSm()->last_snapshot()->get_last_log_idx();
+
+    CHK_TRUE( !(snp_last_log_idx1 == snp_last_log_idx2 && snp_last_log_idx1 == snp_last_log_idx3));
```
greensky00, line 2099: Actually this may cause a false test failure, as random values can be the same by accident. Do you have any good idea to avoid this issue?

Or maybe we can modify like below
```c++
std::uniform_int_distribution<int32>
    distribution( params->snapshot_distance_ / 2, 
                  std::max( params->snapshot_distance_ / 2, 
                            params->snapshot_distance_ - 1 ) );
```
and check 
```c++
CHK_NEQ(NUM, snp_last_log_idx1)
```

src/raft_server.cxx
```diff
@@ -117,6 +118,21 @@ raft_server::raft_server(context* ctx, const init_options& opt)
     if (params->stale_log_gap_ < params->fresh_log_gap_) {
         params->stale_log_gap_ = params->fresh_log_gap_;
     }
+    if (params->enable_randomized_snapshot_creation_ && !get_last_snapshot()) {
```
greensky00, line 121: I think another condition is needed
```c++
params->snapshot_distance_ > 1
```
For 0, the snapshot is disabled, and for 1, there is no meaning of random generation.

JackyWoo: Maybe it is better use 0 for code readability.

greensky00 (approved):
LGTM, thanks!

PR #373, opened 2022-08-29
Fix segfault during snapshot install

antonio2368:
Snapshot sync can call log compaction while committing is done at the same time in the background.
Commit bg thread goes into the loop of committing logs up to a certain log idx but during that, those logs could end up being compacted.

To prevent this, we pause the bg commit thread while sync is in progress.
No other place calls log compaction apart from the commit thread itself.

### DIFF
--- a/src/handle_commit.cxx
+++ b/src/handle_commit.cxx
@@ -215,9 +215,18 @@ bool raft_server::commit_in_bg_exec(size_t timeout_ms) {
         }
 
         ulong index_to_commit = sm_commit_index_ + 1;
-        ptr<log_entry> le = log_store_->entry_at(index_to_commit);
-        p_tr( "commit upto %llu, curruent idx %llu\n",
+        p_tr( "commit upto %llu, current idx %llu\n",
               quick_commit_index_.load(), index_to_commit );
+
+        ptr<log_entry> le = log_store_->entry_at(index_to_commit);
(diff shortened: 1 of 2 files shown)


greensky00 (approved):
@antonio2368 
I understand how your change works. But I wonder why the commit thread was working while the snapshot installation was in progress, actually it shouldn't happen. Can you please elaborate on it a bit more? 

My guess is, the commit thread (`sm_commit_index_`) was lagging behind `quick_commit_index_` so it was catching it up during the snapshot installation. Even at the moment of `is_last_obj = true`, it couldn't finish the catch-up so `sm_commit_index_` was smaller than `quick_commit_index_` and even smaller than `req.get_snapshot().get_last_log_idx()`. That's why it touched the log index that was compacted by the snapshot receiver and caused the crash. Am I correct?

antonio2368:
@greensky00 your explanation is correct. 
700k logs were being committed, and when bg commit thread goes into loop to commit something, it won't stop until all of `quick_commit_index_ - sm_commit_index_` logs are committed.

This is the first time we saw this segfault probably because you need to have a huge amount of logs without snapshot. The use case that produced it had a snapshot distance of 1 million giving enough time for commit thread and snapshot install to overlap.

greensky00:
@antonio2368 Thanks for confirming.

PR #377, opened 2022-09-27
Make manually create snapshot result knowable

JackyWoo:
### Use case:
In my case snapshot creation is asynchronous , we want to know when creation is done.

### Change log:

1. `create_snapshot` return `ulong` which is back forward compatible
2. add API `snapshot_created`

### DIFF
--- a/include/libnuraft/raft_server.hxx
+++ b/include/libnuraft/raft_server.hxx
@@ -743,9 +743,16 @@ public:
      * Manually create a snapshot based on the latest committed
      * log index of the state machine.
      *
-     * @return `true` on success.
+     * @return `true` on success or on scheduled for asynchronously creation.
      */
-    bool create_snapshot();
+    ulong create_snapshot();
+
+    /**
+     * Check whether a snapshot is created
+     * @param log_idx
+     * @return true if last_log_idx_ of last snapshot or equal with or bigger than log_idx
+     */
+    bool snapshot_created(ulong log_idx);
 
 protected:
     typedef std::unordered_map<int32, ptr<peer>>::const_iterator peer_itor;
--- a/include/libnuraft/raft_server.hxx
+++ b/include/libnuraft/raft_server.hxx
@@ -743,16 +743,17 @@ public:
      * Manually create a snapshot based on the latest committed
      * log index of the state machine.
      *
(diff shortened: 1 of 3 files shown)


include/libnuraft/raft_server.hxx
```diff
@@ -743,9 +743,16 @@ public:
      * Manually create a snapshot based on the latest committed
      * log index of the state machine.
      *
-     * @return `true` on success.
+     * @return `true` on success or on scheduled for asynchronously creation.
```
greensky00, line 746: This should be corrected:
```
@return Log index number of the created snapshot.
        `0` if failed.
```

include/libnuraft/raft_server.hxx
```diff
@@ -743,9 +743,16 @@ public:
...
+     */
+    ulong create_snapshot();
+
+    /**
+     * Check whether a snapshot is created
+     * @param log_idx
+     * @return true if last_log_idx_ of last snapshot or equal with or bigger than log_idx
      */
-    bool create_snapshot();
+    bool snapshot_created(ulong log_idx);
```
greensky00, line 755: This API is too ad-hoc. I'd rather add an API returning the log index of the last snapshot:
```c++
/**
 * Get the log index number of the last snapshot.
 *
 * @return Log index number of the last snapshot.
 *         `0` if snapshot does not exist.
 */
ulong get_last_snapshot_idx() const;
```

src/handle_commit.cxx
```diff
@@ -463,10 +463,14 @@ bool raft_server::apply_config_log_entry(ptr<log_entry>& le,
...
-bool raft_server::create_snapshot() {
+ulong raft_server::create_snapshot() {
     uint64_t committed_idx = sm_commit_index_;
     p_in("manually create a snapshot on %lu", committed_idx);
-    return snapshot_and_compact(committed_idx, true);
+    return snapshot_and_compact(committed_idx, true) ? committed_idx : 0;
+}
+
+bool raft_server::snapshot_created(ulong committed_idx) {
+    return get_last_snapshot()->get_last_log_idx() >= committed_idx;
```
greensky00, line 473: It should check whether `get_last_snapshot()` is null or not.

greensky00 (approved):

PR #381, opened 2022-10-26
New API `get_first_log_idx` & `get_first_log_term`

JackyWoo:
### Usage

I want to get Raft log informations, such as `first_log_idx`, `first_log_term`, `last_log_idx `, `last_committed_log_idx `, `last_snapshot_idx ` etc.

PS: We can calculate log count with `first_log_idx` and `last_log_idx `.

### DIFF
--- a/include/libnuraft/raft_server.hxx
+++ b/include/libnuraft/raft_server.hxx
@@ -395,6 +395,22 @@ public:
     { return log_store_->term_at(log_idx); }
 
     /**
+     * Get the term of the first log.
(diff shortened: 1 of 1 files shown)


greensky00 (commented):
These APIs are not needed as they are just shortcuts to existing ways, for example
```c++
raft->get_log_store()->start_index();
```
It will be better to define your own helper functions if you need them.

JackyWoo:
Thanks.

PR #385, opened 2022-11-16
Checking parameters

JackyWoo:
Check election timeout lower and upper bound.

### DIFF
--- a/src/raft_server.cxx
+++ b/src/raft_server.cxx
@@ -372,41 +372,42 @@ void raft_server::update_params(const raft_params& new_params) {
 
 void raft_server::apply_and_log_current_params() {
     ptr<raft_params> params = ctx_->get_params();
-    p_in("parameters: "
-         "timeout %d - %d, heartbeat %d, "
-         "leadership expiry %d, "
-         "max batch %d, backoff %d, snapshot distance %d, "
-         "enable randomized snapshot creation %s, "
-         "log sync stop gap %d, "
-         "reserved logs %d, client timeout %d, "
-         "auto forwarding %s, API call type %s, "
-         "custom commit quorum size %d, "
-         "custom election quorum size %d, "
-         "snapshot receiver %s, "
-         "leadership transfer wait time %d, "
-         "grace period of lagging state machine %d, "
-         "snapshot IO: %s, "
-         "parallel log appending: %s",
-         params->election_timeout_lower_bound_,
-         params->election_timeout_upper_bound_,
-         params->heart_beat_interval_,
-         get_leadership_expiry(),
-         params->max_append_size_,
-         params->rpc_failure_backoff_,
-         params->snapshot_distance_,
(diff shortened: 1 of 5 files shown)


src/raft_server.cxx
```diff
@@ -409,6 +409,12 @@ void raft_server::apply_and_log_current_params() {
           params->use_bg_thread_for_snapshot_io_ ? "ASYNC" : "BLOCKING",
           params->parallel_log_appending_ ? "ON" : "OFF" );
 
+    if (params->election_timeout_lower_bound_ >= params->election_timeout_upper_bound_) {
+        p_ft("Params election_timeout_upper_bound_ should larger than "
+             "election_timeout_lower_bound_");
+        ::exit(-1);
```
greensky00, line 415: Given that parameters can be changed at any time on the fly, it should be handled smoothly rather than killing the process. I'd recommend automatic adjustment of the upper bound:
```c++
if (params->election_timeout_lower_bound_ >= params->election_timeout_upper_bound_) {
    params->election_timeout_upper_bound_ = params->election_timeout_lower_bound_ * 2;
    p_in("invalid upper bound detected, adjusted to XX", ...);
}
```
Note that the same thing should be done for `heartbeat` and `lower_bound` as well (`lower_bound` should be greater than `heartbeat`).

If you really want to kill the process in your case, you can add a callback type (let's say `InvalidParameterDetected`) to notify such invalid parameters:
https://github.com/eBay/NuRaft/blob/1c2587f159775a4924573db64ffac5c973230761/include/libnuraft/callback.hxx#L28
and add
```c++
ctx_->cb_func_.call(cb_func::InvalidParameterDetected, &param);
```
in your callback handler, you can do `exit(-1)`.

JackyWoo: Thanks for advice.

JackyWoo:
Trigger some test errors. I try to fix in #386, pls correct me if there is something wrong.

greensky00:
Election timeouts and heartbeat in that test were set on purpose. That test does not use a "real" timer but artificially coordinates the timing to produce the edge cases that can hardly happen in the real world. 

I'd rather add a flag for test mode in `raft_Server`.
```c++
    /**
     * If `true`, test mode is enabled.
     */
    std::atomic<bool> test_mode_flag_;

    /**
     * Set the test mode.
     * 
     * @param to Test mode.
     */
    void set_test_mode(bool to) {
        test_mode_flag_ = to;
    }
```
and do the auto adjustment only when the test mod is false:
```c++
if (test_mode_flag_) {
    if (params->heart_beat_interval_ >= params->election_timeout_lower_bound_) 
    ...
}
```

Also, enable the flag here:
https://github.com/eBay/NuRaft/blob/1c2587f159775a4924573db64ffac5c973230761/tests/unit/raft_package_fake.hxx#L90

greensky00 (approved):
Thanks!

PR #386, opened 2022-11-17
Fix test error

JackyWoo:
Change log :

1. `election_timeout_lower_bound_` can not be 0.
2. Add `invoke_election_timer` method identicating that we must wait election timeout before trigger leader election.
3. Only `S1` is leader in `init_options_test` but not all .
4. Remove server msg should deliver to `S3` in `rmv_not_resp_srv_wq_test`.
5. We should sleep some time lower than `heart_beat_interval_` when waiting reconnecting in `follower_reconnect_test`.
6. Change `RaftAsioPkg::HEARTBEAT_MS` to 125 (the default value)


greensky00 (changes requested):
As I mentioned https://github.com/eBay/NuRaft/pull/385, you should not touch this artificially controlled timer. Timers used in this test are not real timers. Please close the PR.

PR #391, opened 2023-01-22
chore: remove travis config

chenrui333:
As there is no build status anymore with travis, good to clean it up from the source tree. Let me know if that makes sense.


greensky00 (approved):
Thanks!

PR #394, opened 2023-01-30
do nothing in ssl_test() when build with -DDISABLE_SSL=1

byronhe:


greensky00:
@byronhe Thank you for finding this! Instead of making the test always succeeded, can we skip the test itself like this?
```c++
#if !SSL_LIBRARY_NOT_FOUND && (defined(__linux__) || defined(__APPLE__))
    ts.doTest( "ssl test",
               ssl_test );
#endif
```

PR #395, opened 2023-02-07
1. do nothing in ssl_test() when build with -DDISABLE_SSL=1   2. add printf format check for log

byronhe:
1. do nothing in ssl_test() when build with -DDISABLE_SSL=1 
3. add printf format check for log


greensky00:
Thanks. Since your PR will expose all format-related warnings, let me first fix all those warnings and then merge yours.

greensky00 (approved):

PR #396, opened 2023-02-12
Fix a typo "idnex" -> "index"

alexey-milovidov:
I was not sure about it, but the previous lines suggest "index".


greensky00 (dismissed):
Thanks!

greensky00 (approved):

PR #398, opened 2023-02-18
add asio as module

bisegni:
in this way this project can be used directly as submodule without problem and without calling the ./prepare.sh script

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,6 @@ debug/
 Debug/
 .vs/
 *.user
-asio/
 nbproject/
 *.obj
 *.o
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "asio"]
+	path = asio
+	url = https://github.com/chriskohlhoff/asio.git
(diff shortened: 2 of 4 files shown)


greensky00:
Hi @bisegni , thanks for submitting this PR.

I don't want to break the existing building scripts after merging this change. Since a submodule creates an empty directory in advance, `prepare.sh` may not work. So can you modify [this line](https://github.com/eBay/NuRaft/blob/f6775709b477569fe7c54b58be77dd00b6e9b9f3/prepare.sh#L10) as follows?
```c++
if [ ! -d asio/asio ]; then
```

And please use the official release commit, not the current master. The latest release of Asio is:
https://github.com/chriskohlhoff/asio/commit/147f7225a96d45a2807a64e443177f621844e51c

And also, please add a description of how to set up the submodule in Section 2 of [README](https://github.com/eBay/NuRaft/blob/master/README.md) file.
```
$ git submodule update --init
```

Please let me know if you want me to make the above changes.

bisegni:
Hi @greensky00 I should have fixed and applied all your requests. Let me know

greensky00 (approved):
Thanks!

Issue #400, opened 2023-02-23
integer overflow ?


cbucher:
```cpp
enum log_val_type {
    app_log         = 1,
    conf            = 2,
    cluster_server  = 3,
    log_pack        = 4,
    snp_sync_req    = 5,
    custom          = 999,
};
```
We can see that the "custom" entrie of the log_val_type enum is set to 999. However, when this value is serialized, it
is cast as 8bit unsigned integer which maximum value is 255.

```cpp
    ptr<buffer> serialize() {
        buff_->pos(0);
        ptr<buffer> buf = buffer::alloc( sizeof(ulong) +
                                         sizeof(char) +
                                         buff_->size() );
        buf->put(term_);
        buf->put( (static_cast<byte>(value_type_)) );
        buf->put(*buff_);
        buf->pos(0);
        return buf;
    }
```

Perhaps, the enum could be defined as :
```cpp
enum log_val_type : byte {
    app_log         = 1,
    conf            = 2,
    cluster_server  = 3,
    log_pack        = 4,
    snp_sync_req    = 5,
    custom          = 999,
};
```

greensky00:
Hi @cbucher , thanks for bringing up this issue. Actually it was already reported before https://github.com/eBay/NuRaft/issues/349 and then forgotten.

I'd rather prefer changing it like this:
```c++
enum log_val_type : byte {
    app_log         = 1,
    conf            = 2,
    cluster_server  = 3,
    log_pack        = 4,
    snp_sync_req    = 5,
    custom          = 231,
};
```
given that `999 % 256 = 231`. Would you like to submit a PR? Or let me know if you want me to do it.

cbucher:
Hi,

Yes I will submit a PR.

cbucher:
Hi  @greensky00 , I have submitted a PR #404

Best regards

PR #404, opened 2023-03-03
Fix possible integer overflow when serializing log_val_type.

cbucher:
We can see that the "custom" entrie of the log_val_type enum is set to 999. However, when this value is serialized, it is cast as 8bit unsigned integer which maximum value is 255.

We type the enumeration to 8bit interger and we change the value of custom to 999 (given that 999 % 256 = 231).

#400

### DIFF
--- a/include/libnuraft/log_val_type.hxx
+++ b/include/libnuraft/log_val_type.hxx
@@ -23,13 +23,13 @@ limitations under the License.
 
 namespace nuraft {
 
-enum log_val_type {
+enum log_val_type : byte {
     app_log         = 1,
     conf            = 2,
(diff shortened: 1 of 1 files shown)


greensky00 (approved):
Thanks!

antonio2368:
This actually breaks backward compatibility because of serialization (previously it wasn't written as byte because of `999`).
😢

greensky00:
@antonio2368 sorry for the trouble. There are a few more `enum`s without explicit type, and they are actually treated as `byte` (== `uint8_t`). These should be also fixed. Are you using any of them?

PR #405, opened 2023-03-03
fix Visual Studio build

cbucher:
fix Visual Studio build :
src\tracer.hxx(34): error C3646: '__attribute__': unknown override specifier

We use the Microsoft source-code annotation language (SAL) equivalent to __attribute__((format(printf, ...

### DIFF
--- a/src/tracer.hxx
+++ b/src/tracer.hxx
@@ -24,13 +24,15 @@ limitations under the License.
 
 #include <stdarg.h>
 
-static inline std::string msg_if_given(const char* format, ...)
-    __attribute__((format(printf, 1, 2)));
-
-
+#ifdef _WIN32
+static inline std::string msg_if_given
+                          ( _Printf_format_string_ const char* format,
+                            ... ) {
+#else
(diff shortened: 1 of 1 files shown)


greensky00:
Please take a look at this build error:
https://github.com/eBay/NuRaft/actions/runs/4324359161/jobs/7866385517?pr=405#step:4:46

It seems to me both the declaration and definition should be separated:
```c++
#ifdef _WIN32
static inline std::string msg_if_given
                          ( _Printf_format_string_ const char* format,
                            ... ) {
// ...
#else
static inline std::string msg_if_given(const char* format, ...)
    __attribute__((format(printf, 1, 2)));

static inline std::string msg_if_given
                          ( const char* format,
                            ... ) {
    if (format[0] == 0x0) {
        return "";
    } else {
// ...
#endif
```

cbucher:
That's fixed. You can squash merge.
Regards

greensky00 (commented):
Thanks!

greensky00 (approved):

PR #412, opened 2023-04-12
Allowing clients to provide pre-Configured SSL_CTX for both client and server use-cases.

Enmk:
To avoid adding too many configuration parameters to the asio_service_options, two new callbacks were added:
- std::function<SSL_CTX* (void)> ssl_context_provider_server_;
- std::function<SSL_CTX* (void)> ssl_context_provider_client_;

Which provide preconfigured, ready to use SSL_CTX, so the clients of the library may configure contexts in any required way.

Ported from https://github.com/ClickHouse/NuRaft/pull/56

### DIFF
--- a/include/libnuraft/asio_service_options.hxx
+++ b/include/libnuraft/asio_service_options.hxx
@@ -21,6 +21,9 @@ limitations under the License.
 #include <string>
 #include <system_error>
 
+
+typedef struct ssl_ctx_st SSL_CTX;
+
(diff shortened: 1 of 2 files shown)


greensky00 (commented):
Thanks @Enmk I have a couple of minor comments.

greensky00 (approved):
Since my comments are all trivial changes, let me merge it first and modify them in a separate commit. Thanks.

PR #413, opened 2023-04-15
Allow get_servers() on const cluster_config

szmyd:
Currently one must const_cast a cluster_config in order to iterate the servers it contains. This is done potentially during commit_conf or snapshots if one does not use the serialize/deserialize methods.

### DIFF
--- a/include/libnuraft/cluster_config.hxx
+++ b/include/libnuraft/cluster_config.hxx
@@ -69,7 +69,7 @@ public:
         return prev_log_idx_;
     }
 
-    std::list<ptr<srv_config>> const& get_servers() const {
+    const std::list<ptr<srv_config>>& get_servers() const {
         return servers_;
     }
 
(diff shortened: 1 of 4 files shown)


include/libnuraft/cluster_config.hxx
```diff
@@ -72,6 +69,10 @@ public:
         return prev_log_idx_;
     }
 
+    std::list<ptr<srv_config>> const& get_servers() const {
```
greensky00, line 72: Can you please use
```c++
const std::list<ptr<srv_config>>&
```
to make coding style aligned

src/raft_server.cxx
```diff
@@ -196,8 +196,8 @@ raft_server::raft_server(context* ctx, const init_options& opt)
                               log_store_->start_index() );
           i < log_store_->next_slot();
           ++i ) {
-        ptr<log_entry> entry(log_store_->entry_at(i));
-        if (entry->get_val_type() == log_val_type::conf) {
+        if (auto entry = log_store_->entry_at(i);
+                              entry->get_val_type() == log_val_type::conf) {
```
greensky00, line 200: This is supported starting from c++17, may need to use the old way.

greensky00 (approved):

PR #432, opened 2023-05-25
Return status of set_priority. Ability to broadcast priority changes if there's a leader

myrrc:
`set_priority` is a bit inconsistent with `add_srv` and `remove_srv`.
If you invoke the latter two on a follower, the request will be
forwarded to leader. You can wait on the request and check whether it was accepted or not. 
Unfortunately, there are no such guarantees for `set_priority`.
If you invoke it on a follower, and there is a live leader, function would quit. 
I want to know the result of `set_priority`, so changes are proposed that change its return type.

### DIFF
--- a/include/libnuraft/raft_server.hxx
+++ b/include/libnuraft/raft_server.hxx
@@ -314,15 +314,14 @@ public:
         append_entries_ext(const std::vector< ptr<buffer> >& logs,
                            const req_ext_params& ext_params);
 
-    enum class priority_set_result {
-        SET = 0,
-        BROADCAST = 1,
-        IGNORED = 2
+    enum class PrioritySetResult {
+        SET,
+        BROADCAST,
+        IGNORED
     };
 
     /**
      * Update the priority of given server.
-     * Only leader will accept this operation.
      *
      * @param srv_id ID of server to update priority.
      * @param new_priority
@@ -334,11 +333,11 @@ public:
      * @return SET If we're a leader and we have committed priority change.
      * @return BROADCAST
      *     If either there's no live leader now, or we're a leader and we want to set our priority to 0,
-     *     or broadcast_when_leader_exists = true and we're not a leader.
+     *     or we're not a leader and broadcast_when_leader_exists = true.
      *     We have sent messages to other peers about priority change but haven't committed this change.
      * @return IGNORED If we're not a leader and broadcast_when_leader_exists = false. We ignored the request.
      */
-    priority_set_result set_priority(
+    PrioritySetResult set_priority(
         const int srv_id, const int new_priority,
         bool broadcast_when_leader_exists = false);
 
--- a/include/libnuraft/raft_server.hxx
+++ b/include/libnuraft/raft_server.hxx
@@ -314,11 +314,7 @@ public:
         append_entries_ext(const std::vector< ptr<buffer> >& logs,
                            const req_ext_params& ext_params);
 
-    enum class PrioritySetResult {
-        SET,
-        BROADCAST,
-        IGNORED
-    };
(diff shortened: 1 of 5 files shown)


include/libnuraft/raft_server.hxx
```diff
@@ -314,16 +314,32 @@ public:
...
-    void set_priority(const int srv_id, const int new_priority);
+     * @param broadcast_when_leader_exists
+     *     If we're not a leader and a leader exists, broadcast priority change to other peers.
+     *     If false, set_priority does nothing.
+     * @return SET If we're a leader and we have committed priority change.
+     * @return BROADCAST
+     *     If either there's no live leader now, or we're a leader and we want to set our priority to 0,
+     *     or we're not a leader and broadcast_when_leader_exists = true.
+     *     We have sent messages to other peers about priority change but haven't committed this change.
+     * @return IGNORED If we're not a leader and broadcast_when_leader_exists = false. We ignored the request.
```
greensky00, line 338: Please break lines at 90 chars. 

Also, in the explanation of `broadcast_when_leader_exists`, we should mention that it will cause `cluster_config` to diverge, so setting this parameter to `true` is at the user's own risk.

tests/unit/asio_service_test.cxx
```diff
@@ -1245,11 +1245,11 @@ int leadership_transfer_test() {
     CHK_EQ(1, s3->raftServer->get_leader());
 
     // Set the priority of S2 to 10.
-    s1->raftServer->set_priority(2, 10);
+    CHK_TRUE( s1->raftServer->set_priority(2, 10) == raft_server::PrioritySetResult::SET );
```
greensky00, line 1248: It exceeds 90 chars. Please replace these with
```c++
CHK_EQ( raft_server::PrioritySetResult::SET, s1.raftServer->set_priority(2, 100) );
```
like you did in the other file.

greensky00 (commented):
Hi @myrrc, thanks for submitting this PR.

I understand how it works, and the overall code looks good to me, but can you please elaborate on why (and in which case) this option is needed?

Also, please break lines at 90 chars. There are a few places to fix.

myrrc:
Hi @greensky00.

This PR is needed for my other PR regarding a ZooKeeper-like program. Each ZooKeeper-like replica
has a NuRaft server inside, and it can be sent ZooKeeper-compatible commands. When a command gets
to the replica, replica must process it or abort. For `add/remove_srv` we can check whether result was
accepted. For `set_priority` we don't have this guarantee.

What should we do if we don't want to lose the priority update message? We can't rely on leadership and
only process message at the current leader as:
- there is a chance there will be elections after returning from `is_leader()` and before appending new config
  to log in `set_priority` in which case we'll lose the update.
- there is a counter-example on {A, B, C} cluster where command going to A sees B as leader,
  then there are re-elections and command on B sees C as leader,
  re-elections again, and then command on C sees A as leader.
  In this case changes would also be lost.

Nor can we simply change our config via `get_server(id)->set_priority` as NuRaft has various invariants inside
like `uncommitted_config_` that would be broken.
So, we need to `set_priority` on each replica explicitly. However, if there's a live leader,
unlike `add/remove_srv`, the new config won't be forwarded (I tried implementing this but failed as it
seems quite a big task) and we'll lose the update.

Thus, a new option is introduced that would send messages to replicas telling to update priority even when a
leader is present. In my case there are no chances of going out of sync (as update will be propagated to every
replica) but I agree it can happen in general. Please note, however, that invoking `set_priority` when
there's no live leader currently can also lead to inconsistencies as a peer can reject the priority update
message (`is_bisy()` case).

I will address formatting issues and documentation for `set_priority` today.

Kind regards,
Mike.

greensky00 (approved):
Thanks.

PR #441, opened 2023-06-12
Extract the public interface class global_mgr, so users can provide their own global_mgr implement.

byronhe:
Extract the public interface class global_mgr, so users can provide their own global_mgr implement.
so in multi raft instance use scenes, users can do their own performance optimization or multi-global_mgr-in-one-process test.

### DIFF
--- a/include/libnuraft/context.hxx
+++ b/include/libnuraft/context.hxx
@@ -25,6 +25,7 @@ limitations under the License.
 #include "pp_util.hxx"
 #include "ptr.hxx"
 #include "raft_params.hxx"
+#include "global_mgr.hxx"
 
 #include <memory>
 #include <mutex>
@@ -37,6 +38,7 @@ class rpc_client_factory;
 class rpc_listener;
 class state_machine;
 class state_mgr;
+class global_mgr;
 struct context {
 public:
     context( ptr<state_mgr>& mgr,
@@ -88,6 +90,13 @@ public:
         params_ = to;
     }
 
+    global_mgr * get_global_mgr() const {
+        if (custom_global_mgr_ != nullptr) {
+            return custom_global_mgr_;
+        }
+        return nuraft_global_mgr::get_instance();
+    }
+
     __nocopy__(context);
 
 public:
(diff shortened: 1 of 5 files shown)


greensky00:
+)
Please let me know if what you intended was selectively using different global managers for different Raft instances running in the same process. If that's the case, relying on `ctx_` makes sense, and the next discussion needs to be based on that way.

byronhe:
> +)
> Please let me know if what you intended was selectively using different global managers for different Raft instances running in the same process. If that's the case, relying on `ctx_` makes sense, and the next discussion needs to be based on that way.

yes, in our use case, we want assign different global manager for different raft instance, in the same process.

include/libnuraft/context.hxx
```diff
@@ -88,6 +90,13 @@ public:
         params_ = to;
     }
 
+    global_mgr * get_global_mgr() const {
+        if (custom_global_mgr_ != nullptr) {
+            return custom_global_mgr_;
+        }
+        return nuraft_global_mgr::get_instance();
+    }
+
```
greensky00, line 99: `nuraft_global_mgr` shouldn't be involved in `context` and also `context` shouldn't be aware of the singleton. Please move this function out and put it into `raft_server`:
```c++
global_mgr* raft_server::get_global_mgr() const {
    if (ctx_->custom_global_mgr_) {
        return ctx_->custom_global_mgr_;
    }
    return nuraft_global_mgr::get_instance();
}
```

include/libnuraft/context.hxx
```diff
@@ -131,6 +140,13 @@ public:
      */
     cb_func cb_func_;
 
+    /**
+     *  User-provided global_mgr pointer.
+     *  used in preference to nuraft_global_mgr::get_instance().
```
greensky00, line 145: Used

include/libnuraft/context.hxx
```diff
@@ -131,6 +140,13 @@ public:
...
+     *  The lifecycle of this object must be managed by the user externally.
+     */
+    global_mgr* custom_global_mgr_{nullptr};
```
greensky00, line 148: Please make initialization style aligned:
```c++
global_mgr* custom_global_mgr_;
```
and in constructor
```c++
        , params_( cs_new<raft_params>(params) )
        , custom_global_mgr_(nullptr)
        {}
```

greensky00 (approved):
Thanks, fixed one typo in the comment.

PR #446, opened 2023-07-12
fix : raft_server::~raft_server should not block wait when using global_mgr

byronhe:


greensky00 (approved):
Thanks, I did link breaking.

PR #454, opened 2023-08-07
Add optional CRC per log entry.

yong-li:
Add an option to replicate and verify per-log-entry CRC checksums.

The log entry's checksum is computed when the entry is first created on the leader, and then stored in the log store and replicated over the network. On receive, if the checksum is sent, it is verified.

### DIFF
--- a/include/libnuraft/asio_service_options.hxx
+++ b/include/libnuraft/asio_service_options.hxx
@@ -123,6 +123,7 @@ struct asio_service_options {
         , custom_resolver_(nullptr)
         , replicate_log_timestamp_(false)
         , crc_on_entire_message_(false)
+        , crc_on_payload_(false)
         , corrupted_msg_handler_(nullptr)
         {}
 
@@ -243,7 +244,7 @@ struct asio_service_options {
      * restore the timestamp when it reads log entries.
      *
      * This feature is not backward compatible. To enable this feature, there
-     * should not be any member running with old version before supprting
+     * should not be any member running with old version before supporting
      * this flag.
      */
     bool replicate_log_timestamp_;
@@ -257,7 +258,7 @@ struct asio_service_options {
     /**
      * If `true`, each log entry will contain a CRC checksum of the entry's
      * payload.
-     * 
+     *
      * To support this feature, the log store implementation should be able to
      * store and retrieve the CRC checksum when it reads log entries.
      *
--- a/include/libnuraft/asio_service_options.hxx
+++ b/include/libnuraft/asio_service_options.hxx
@@ -255,6 +255,19 @@ struct asio_service_options {
     bool crc_on_entire_message_;
 
     /**
+     * If `true`, each log entry will contain a CRC checksum of the entry's
+     * payload.
+     * 
+     * To support this feature, the log store implementation should be able to
+     * store and retrieve the CRC checksum when it reads log entries.
+     *
+     * This feature is not backward compatible. To enable this feature, there
+     * should not be any member running with the old version before supporting
+     * this flag.
+     */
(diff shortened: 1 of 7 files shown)


include/libnuraft/log_entry.hxx
```diff
@@ -92,6 +103,14 @@ public:
         timestamp_us_ = t;
     }
 
+    uint32_t get_crc32() const {
+        return crc32_;
+    }
+
+    void set_crc32(uint32_t crc) {
+        crc32_ = crc;
+    }
```
greensky00, line 112: For backward compatibility (for the previous logs that do not have CRC, and they can be mixed with new logs with CRC), we should have one more field indicating if CRC exists or not, let's say `has_crc32_`. 

Otherwise, we cannot distinguish non-existing CRC and CRC=0. Also we need an API returning that flag: 
```c++
bool has_crc32() const () { ... }
```
that will be used in asio layer (replacing `crc32 != 0`).

src/asio_service.cxx
```diff
@@ -1226,6 +1256,9 @@ class asio_rpc_client
             if (impl_->get_options().replicate_log_timestamp_) {
                 ss.put_u64( le->get_timestamp() );
             }
+            if (impl_->get_options().crc_on_payload_) {
+                ss.put_u32(le->get_crc32());
+            }
```
greensky00, line 1261: Similarly, we need to ship the existence flag as well.
```c++
if (impl_->get_options().crc_on_payload_) {
    ss.put_u8(le->has_crc32() ? 1 : 0);
    ss.put_u32(le->get_crc32());
}
```
and
```c++
LOG_ENTRY_SIZE += 5;
```
I don't have better idea rather than shipping 5 bytes.

include/libnuraft/log_entry.hxx
```diff
@@ -38,12 +40,21 @@ public:
     log_entry(ulong term,
               const ptr<buffer>& buff,
               log_val_type value_type = log_val_type::app_log,
-              uint64_t log_timestamp = 0)
+              uint64_t log_timestamp = 0,
+              uint32_t crc32 = 0,
+              bool compute_crc = true)
```
greensky00, line 45: Can this be `false` by default?

yong-li: This is to cope with conf logs.  There are more than 1 place where conf logs are constructed. Either we need to copy the CRC computation code (or this true flag) everywhere (and I need to make sure I found every place), or we bite the performance overhead and lean towards safety.

include/libnuraft/log_entry.hxx
```diff
@@ -23,8 +23,10 @@ limitations under the License.
 
 #include "basic_types.hxx"
 #include "buffer.hxx"
+#include "crc32.hxx"
```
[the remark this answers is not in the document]
yong-li: Alright. Will do.

include/libnuraft/asio_service_options.hxx
```diff
@@ -254,6 +254,19 @@ struct asio_service_options {
...
+     * payload.
+     * 
+     * To support this feature, the log store implementation should be able to
+     * store and retrieve the CRC checksum when it reads log entries.
+     *
+     * This feature is not backward compatible. To enable this feature, there
+     * should not be any member running with the old version before supporting
+     * this flag.
+     */
+    bool crc_on_payload_;
```
greensky00, line 268: Initialization in the constructor is missing.

greensky00 (approved):
LGTM, but one missing initialization. Let me add the fix and merge it. Thanks.

greensky00 (approved):

PR #461, opened 2023-08-22
Add missing alias targets to CMakeLists

### LINKED ISSUE #459: CMake targets should be namespaced using ALIAS targets
There are multiple ways to import a CMake project, so it's important that usage is consistent. This is why it's common to use ALIAS targets to namespace targets in projects integrated with tools such as FetchContent. See #458.

JosiahWI:
This makes usage consistent. NuRaft::target_name will always work regardless of whether NuRaft was set up as a subdirectory or loaded from a config.

Fixes #459.


greensky00 (approved):
Thanks.

PR #470, opened 2023-09-06
ServerJoinFailed callback

myrrc:
If we `add_srv`, what we can check is that a command was accepted. However, the only way to check whether it was applied is to query replica's config. In case of server not being added (e.g. non-existent host) we will never find this server in a config.
This PR proposes a way to handle such corner cases without relying on some arbitrary timeouts

Note we don't need a ServerLeaveFailed callback as a server that fails to be removed gracefully is still removed.

### DIFF
--- a/src/handle_join_leave.cxx
+++ b/src/handle_join_leave.cxx
@@ -95,6 +95,10 @@ ptr<resp_msg> raft_server::handle_add_srv_req(req_msg& req) {
         // Otherwise: activity timeout, reset the server.
         p_wn("activity timeout (last activity %" PRIu64 " ms ago), start over",
              last_active_ms);
+
+        cb_func::Param param(id_, leader_, srv_to_join_->get_id());
+        invoke_callback(cb_func::ServerJoinFailed, &param);
+
         reset_srv_to_join();
     }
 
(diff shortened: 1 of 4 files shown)


src/handle_join_leave.cxx
```diff
@@ -596,6 +596,9 @@ void raft_server::handle_join_leave_rpc_err(msg_type t_msg, ptr<peer> p) {
               p->get_id() );
         config_changing_ = false;
         reset_srv_to_join();
+
+        cb_func::Param param(id_, leader_, p->get_id());
+        invoke_callback(cb_func::ServerJoinFailed, &param);
```
greensky00, line 601: There is one more place to call it:
https://github.com/eBay/NuRaft/blob/99eeef34a2620686e0dd40ad7fbd5cab561140fc/src/handle_join_leave.cxx#L96-L98
Not only by explicit error (this function), but also by timeout.

greensky00 (commented):
Thanks, can you please take a look into my comment?

src/handle_join_leave.cxx
myrrc: Done, thanks

greensky00 (approved):
LGTM

PR #477, opened 2023-11-07
Add a new API to reset the whole state machine.

JackyWoo:
### Use case

If a node leave cluster for a long time, when it startup it wil first load local data into memory then it will receive the `append_entry` request from leader and leader will check data ga. If the gap can not satisfied by log, leader will send sanpshot.The newly startup node reveive the sanpshot and load it into memory.

But it should clear data in memory first, if not inconsistence may occur, details see [here](https://github.com/JDRaftKeeper/RaftKeeper/pull/100).

I think it is better to add a new reset API.

### DIFF
--- a/examples/calculator/calc_state_machine.hxx
+++ b/examples/calculator/calc_state_machine.hxx
@@ -179,6 +179,8 @@ public:
 
         ptr<snapshot_ctx> ctx = entry->second;
         cur_value_ = ctx->value_;
+        // Now the last committed_idx is the snapshot's last log index.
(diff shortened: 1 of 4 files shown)


greensky00:
@JackyWoo 
Actually, `apply_snapshot()` is for the purpose of resetting state machine (and then applying snapshot). Why don't you put your clean-up logic into `apply_snapshot()`? Unless `reset()` is independently used for somewhere else in NuRaft, having that API separate from `apply_snapshot()` seems to me redundant.

JackyWoo:
Got it, thanks.

PR #490, opened 2024-03-06
Add change buffer API to log_entry

hkadayam:


tests/unit/serialization_test.cxx
```diff
@@ -207,17 +220,25 @@ int log_entry_test() {
                              data,
                              static_cast<log_val_type>(1 + rnd() % 5) );
     ptr<buffer> buf2 = entry->serialize();
+    if (compare_log_entries(entry, log_entry::deserialize(*buf2)) == -1) {
+        return -1;
+    }
```
greensky00, line 225: I think we can just do
```c++
CHK_Z(compare_log_entries(entry, log_entry::deserialize(*buf2)));
```
isn't it? If so, let me revise it.

greensky00 (commented):
Looks good, just refactoring.

greensky00 (approved):

PR #498, opened 2024-04-02
auto_fwd_resp_handler should forward resp->get_result_code (#497)

byronhe:


greensky00 (commented):
Thanks @byronhe for addressing this issue.

In addition to your fix, there is one more bug here: `code` should be set to `FAILED` when `err` exists:
```c++
    if (err) {
        perr = err;
        code = cmd_result_code::FAILED;
    } else {
        if (resp->get_accepted()) {
            resp_ctx = resp->get_ctx();
            presult->accept();
        }
        code = resp->get_result_code();
    }
```
If you don't mind, can you please fix this as well?

byronhe:
> Thanks @byronhe for addressing this issue.
> 
> In addition to your fix, there is one more bug here: `code` should be set to `FAILED` when `err` exists:
> 
> ```c++
>     if (err) {
>         perr = err;
>         code = cmd_result_code::FAILED;
>     } else {
>         if (resp->get_accepted()) {
>             resp_ctx = resp->get_ctx();
>             presult->accept();
>         }
>         code = resp->get_result_code();
>     }
> ```
> 
> Also `code` should be set regardless of `resp->get_accepted()`.
> 
> If you don't mind, can you please fix this as well?


Thank you for your suggestion. I made fix as your advice. Could you please review it again?

greensky00 (approved):
Thanks!

Issue #502, opened 2024-04-22
src/tracer.hxx:52  stack-buffer-overflow  when using vsnprintf's return value  


byronhe:
![image](https://github.com/eBay/NuRaft/assets/1394332/aef903a9-103d-46a8-bbcd-9f1424be92c7)
![image](https://github.com/eBay/NuRaft/assets/1394332/9df5f467-0dbe-4d5e-837c-1096a91c3d86)

 
at src/tracer.hxx:52

```cpp
 48         len = vsnprintf(msg, 2048, format, args);
 49         va_end(args);
 50 
 51         // Get rid of newline at the end.
 52         if (msg[len-1] == '\n') {
 53             len--;
 54             msg[len] = 0x0;
 55         }
```
when input format and args will produce a long line longer than 2048 bytes ,  vsnprintf's return value `len` is 
> the number of characters which would have been written to the final string if enough space had been available. 

so line 52  `msg[len] = 0x0;`  caused a  `stack-buffer-overflow`

https://man7.org/linux/man-pages/man3/printf.3.html#RETURN_VALUE

> RETURN VALUE
>        Upon successful return, these functions return the number of characters printed (excluding the null byte used to end output to strings).
>        The functions snprintf() and vsnprintf() do not write more than size bytes (including the terminating null byte ('\0')).  If the output was truncated due to this
>        limit,  then the return value is the number of characters (excluding the terminating null byte) which would have been written to the final string if enough space
>        had been available.  Thus, a return value of size or more means that the output was truncated.  (See also below under NOTES.)
>        If an output error is encountered, a negative value is returned.
>

PR #503, opened 2024-04-22
fix src/tracer.hxx stack-buffer-overflow when using vsnprintf's return value (#502)

byronhe:


greensky00 (commented):
Thanks, looks good to me. A couple of minor comments.

byronhe:
> Thanks, looks good to me. A couple of minor comments.


Thank you for your suggestion. I made fix according to your advice. Could you please review it again?

greensky00 (approved):
Thanks!

Issue #233, opened 2021-07-19
Buggy cluster initialization


ZaychukAleksey:
Hello. I found 2, IMO, buggy examples.

Run `echo_server` example in different terminals nodes:

```sh
./echo_server 1 localhost:10001
./echo_server 2 localhost:10002
./echo_server 3 localhost:10003
```

<details>
<summary>Example 1</summary>

Then add node 2 to the first server and to the third:

```sh
# first terminal
calc 1> add 2 localhost:10002
```

```sh
# third terminal
calc 3> add 2 localhost:10002
```

Ok. Let's see the list of nodes on each server:

```sh
calc 1> list
server id 1: localhost:10001 (LEADER)
server id 2: localhost:10002

#--
calc 2> list
server id 1: localhost:10001 (LEADER)
server id 2: localhost:10002

#--
calc 3> list
server id 3: localhost:10003 (LEADER)
server id 2: localhost:10002
```

How can it be, that node 2 is listed now in, actually, two clusters with different leaders?

Note, that node 2 really follows node 1, it ignores logs from node 3.

If I shut down node 1, some time later node 2 starts following node 3, although the following command doesn't even show node 3 and the current leader:

```sh
calc 2> list
server id 1: localhost:10001
server id 2: localhost:10002
```

Another funny thing occurs when I restart node 1 and add node 2 to it (because after the restart, it thinks it's the only leader node in the cluster). Node 2 still continue receiving logs from node 3 only, but it thinks that node 1 is now a leader:

```sh
calc 2> list
server id 1: localhost:10001 (LEADER)
server id 2: localhost:10002
```

</details>

<details>
<summary>Example 2</summary>

Add node 2 to node 1, then add node 1 to node 3.
```sh
# first terminal
calc 1> add 2 localhost:10002
```

```sh
# third terminal
calc 3> add 1 localhost:10001
```

Now the picture is as follows:

```sh
calc 1> list
server id 1: localhost:10001
server id 2: localhost:10002

calc 2> list
server id 1: localhost:10001
server id 2: localhost:10002

calc 3> list
server id 3: localhost:10003 (LEADER)
server id 1: localhost:10001
```

Node 1 accepts logs from node 3 although it doesn't even report that node 3 exists. Node 2 accepts nothing, it's just dummy.

</details>

[...]

PR #504, opened 2024-04-25
<fix>: prevent a server joining another cluster whlit it is already in one.

fwhdzh:
This a pull request for fixing the problem mentioned in issue #233 
I've added an additional check to prevent a node from joining multiple clusters.


greensky00 (approved):
Thanks!

PR #515, opened 2024-06-17
Add result code to response messages

### LINKED ISSUE #513: Misleading append_entries result code
Hello,

I ran into an issue with NuRaft v2.1.0 where after calling `append_entries` in auto forwarding mode, my callback is called with `accepted=1` and `result_code=0` but I cannot find my data.

Reproducing
===
Here's how I can easily reproduce:
Parameters:
- election_timeout_lower_bound = 250ms
- election_timeout_upper_bound = 500ms
- heartbeat_interval = 50ms
- leadership_expiry = 200ms
- auto forwarding ON

I start 5 servers. For tests purposes, I run 2 servers in a network namespace netns1 and 3 servers in another one netns2. The two netns are connected with a veth interface. I use netem to add some delay between the two netns.

To reproduce this bug, the leader must be one of the 2 servers in netns1.
I call append_entries with some data on the follower in netns1.
Then I bring down the veth interface between the two network namespaces after waiting some time shorter than the configured netem delay.

I'm expecting to see an error in the cmd_result I get in my callback for append_results on the follower. But it's not the case.
Then I can not find my data anywhere on any server.

Analysis
===
I dug into the code and here's my understanding:
1) The follower forwards a `client_request` to the leader in netns1.
2) Then the leader sends the request to every follower in `handle_cli_request`. It marks the request as accepted.
3) The leader's leadership expires. It cancels all pending requests in `drop_all_pending_commit_elems`
4) The leader sets the result code in the response object as CANCELLED.
5) The leader crafts the response message for the client_request to send back to the follower. The result code is not part of the encoded response message, so only accepted is set to 1.
6) The follower receives and decodes the response message. It creates a resp_msg object, sets accepted to 1 but uses the default result_code in the constructor, i.e. `cmd_result_code::OK`.

Here are the logs of the leader:
[logs.txt](https://github.com/user-attachments/files/15788542/logs.txt)

Proposed solution
===
Here's a patch that I made that seems to solve the issue (note that I'm using version v2.1.0 so I didn't see the patch d195630b):
[patch.txt](https://github.com/user-attachments/files/15788822/patch.txt)

amichon-kalray:
In some cases, the response is marked as accepted even if there was an error. Therefore, including the accepted tag in the response message is not enough: a follower may believe that an auto forwarded request was successful when it wasn't.

Resolves #513

### DIFF
--- a/src/asio_service.cxx
+++ b/src/asio_service.cxx
@@ -738,7 +738,7 @@ private:
        try {
         ptr<buffer> resp_ctx = resp->get_ctx();
         int32 resp_ctx_size = (resp_ctx) ? resp_ctx->size() : 0;
-        int32 result_code_size = sizeof(cmd_result_code);
+        int32 result_code_size = sizeof(int32_t);
 
         uint32_t flags = 0x0;
         size_t resp_meta_size = 0;
--- a/src/asio_service.cxx
+++ b/src/asio_service.cxx
@@ -124,6 +124,9 @@ limitations under the License.
 // If set, each log entry will contain a CRC on the payload.
 #define CRC_ON_PAYLOAD (0x10)
 
+// If set, RPC message (response) includes result code
+#define INCLUDE_RESULT_CODE (0x20)
+
 // =======================
 
 namespace nuraft {
@@ -735,6 +738,7 @@ private:
        try {
         ptr<buffer> resp_ctx = resp->get_ctx();
         int32 resp_ctx_size = (resp_ctx) ? resp_ctx->size() : 0;
+        int32 result_code_size = sizeof(cmd_result_code);
 
         uint32_t flags = 0x0;
         size_t resp_meta_size = 0;
@@ -758,6 +762,13 @@ private:
         }
 
         size_t carried_data_size = resp_meta_size + resp_hint_size + resp_ctx_size;
+
+        if (req->get_type() == msg_type::client_request ||
+            req->get_type() == msg_type::add_server_request ||
+            req->get_type() == msg_type::remove_server_request) {
+            flags |= INCLUDE_RESULT_CODE;
+            carried_data_size += result_code_size;
+        }
 
         int buf_size = RPC_RESP_HEADER_SIZE + carried_data_size;
         ptr<buffer> resp_buf = buffer::alloc(buf_size);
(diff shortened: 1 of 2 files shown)


src/asio_service.cxx
```diff
@@ -1729,9 +1737,21 @@ class asio_rpc_client
         assert(remaining_len >= 0);
         if (remaining_len) {
             // It has context, read it.
-            ptr<buffer> actual_ctx = buffer::alloc(remaining_len);
+            size_t ctx_len = remaining_len;
+            if (flags & INCLUDE_RESULT_CODE) {
+                ctx_len -= sizeof(int32_t);
+                ptr<buffer> actual_ctx = buffer::alloc(ctx_len);
```
greensky00, line 1743: Build fails at here. `ptr<buffer> actual_ctx` should be outside of this scope.

src/asio_service.cxx
```diff
@@ -735,8 +738,9 @@ class rpc_session
        try {
         ptr<buffer> resp_ctx = resp->get_ctx();
         int32 resp_ctx_size = (resp_ctx) ? resp_ctx->size() : 0;
+        int32 result_code_size = sizeof(int32_t);
```
greensky00, line 741: Instead of this, can you please use
```c++
int32 result_code_size = sizeof(cmd_result_code);
```

greensky00 (commented):
Thanks for submitting PR, I put a few comments.

src/asio_service.cxx
```diff
@@ -735,8 +738,9 @@ class rpc_session
...
 
-        uint32_t flags = 0x0;
+        uint32_t flags = INCLUDE_RESULT_CODE;
```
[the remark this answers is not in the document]
amichon-kalray: sorry, didn't pay attention with the rebase on master

src/asio_service.cxx
```diff
@@ -124,6 +124,9 @@ limitations under the License.
 // If set, each log entry will contain a CRC on the payload.
 #define CRC_ON_PAYLOAD (0x10)
 
+// If set, RPC message (response) includes result code
+#define INCLUDE_RESULT_CODE (0x8)
```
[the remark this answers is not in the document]
amichon-kalray: sorry, didn't pay attention with the rebase on master

src/asio_service.cxx
amichon-kalray: ok
my concern was that we will use the put_i32 and get_i32 functions, so it result_code_size is not sizeof(int32_t), weird things may happen

greensky00: Yeah, good point, that makes more sense. Let me make a change to use `sizeof(int32_t)` and instead put `int32_t` to `cmd_result_code` definition.

greensky00 (approved):
LGTM, Thanks.

PR #516, opened 2024-06-17
Add timeout to async_connect operation

amichon-kalray:
If the network goes down during a async_connect, we may end up being stuck forever in this function since there is no timeout to cancel it.


greensky00 (dismissed):
Thanks, let me do some formatting and then merge.

greensky00 (approved):

PR #521, opened 2024-07-18
Matching format specifies type to argument type to eliminate compile …

ZexiLiu:
…warnings

format specifies type 'unsigned long' but the argument has type 'ulong' (aka 'unsigned long long').

### DIFF
--- a/src/raft_server.cxx
+++ b/src/raft_server.cxx
@@ -1198,7 +1198,7 @@ void raft_server::check_leadership_transfer() {
         uint64_t last_resp_ms = peer_elem->get_resp_timer_us() / 1000;
         if (last_resp_ms > election_lower) {
             // This replica is not responding.
-            p_tr("peer %d is not responding, %llu ms ago",
+            p_tr("peer %d is not responding, %" PRIu64 " ms ago",
                  s_conf.get_id(), last_resp_ms);
             return;
         }
--- a/src/raft_server.cxx
+++ b/src/raft_server.cxx
@@ -1189,7 +1189,7 @@ void raft_server::check_leadership_transfer() {
         if (peer_elem->get_matched_idx() + params->stale_log_gap_ <
                 cur_commit_idx) {
             // This peer is lagging behind.
-            p_tr("peer %d is lagging behind, %lu < %lu",
+            p_tr("peer %d is lagging behind, %llu < %llu",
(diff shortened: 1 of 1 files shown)


src/raft_server.cxx
```diff
@@ -1151,7 +1151,7 @@ void raft_server::check_leadership_transfer() {
         if (peer_elem->get_matched_idx() + params->stale_log_gap_ <
                 cur_commit_idx) {
             // This peer is lagging behind.
-            p_tr("peer %d is lagging behind, %lu < %lu",
+            p_tr("peer %d is lagging behind, %llu < %llu",
```
greensky00, line 1154: Please use `PRIu64`. This symbol is different according to the platform.

ZexiLiu: There are many %lu, %llu, %d use case across the whole repo, but I am only fixing this line because it has warning during compile. 

I fixed it to use PRIu64.

greensky00: `%d` probably be ok, but 64-bit integers should be replaced. The whole projected was re-visited a couple of years ago, but if there are still missing ones, we need to fix them.

src/raft_server.cxx
```diff
@@ -1160,7 +1160,7 @@ void raft_server::check_leadership_transfer() {
         uint64_t last_resp_ms = peer_elem->get_resp_timer_us() / 1000;
         if (last_resp_ms > election_lower) {
             // This replica is not responding.
-            p_tr("peer %d is not responding, %lu ms ago",
+            p_tr("peer %d is not responding, %llu ms ago",
```
greensky00, line 1163: Please replace this one too.

src/raft_server.cxx
ZexiLiu:
```
~ grep -rn "%l" .                                                                               
./fix_format.hxx:20:    replace_all(format_string, "%ld", "%" PRId64); \
./fix_format.hxx:21:    replace_all(format_string, "%lu", "%" PRIu64); \
./asio_service.cxx:336:                      "due to error %d, %s, ref count %ld",
```

We don't need to care about ./fix_format.hxx, the %ld in ./asio_service.cxx is correct, I think that is all the case.

src/raft_server.cxx
ZexiLiu: Fixed.

greensky00 (approved):

Issue #522, opened 2024-07-23
Leadership expiry not respected as a limit for leadership expiry


antoniofilipovic:
Hi,

I expected the `leadership_expiry_` flag to be used as the amount of time needed to pass between now and the last time the follower responded to the leader, for the leader to yield its leadership.

However, looking at the code, leadership only expires after followers are deemed unhealthy. And followers are deemed unhealthy according to the following policy:

```
int expiry = params->heart_beat_interval_ *
                     raft_server::raft_limits_.response_limit_;

    // Check the number of not responding peers.
    for (auto& entry: peers_) {
        ptr<peer> p = entry.second;

        if (!is_regular_member(p)) continue;

        int32 resp_elapsed_ms = (int32)(p->get_resp_timer_us() / 1000);
        if ( resp_elapsed_ms > expiry ) {
            num_not_resp_nodes++;
        }
    }

```

I would like to open PR to fix this issue that the expiry used in this function is the same as the leadership expiry time, as otherwise this flag is not used to calculate when will leadership expire.

Let me know if this is actually not an issue, and I didn't understand something properly.

many thanks,
Antonio

greensky00:
Thanks for pointing this out. Looks like this part of code was previously correct, and then the bug has been introduced since the below commit.
https://github.com/eBay/NuRaft/commit/bb863abda5f0e245a680647c0855743b80631c44#diff-60ca73a04b9bb2ab5fa615dc2a38a817a8260975d6c06d35ca90a02555731635R573

As `get_not_responding_peers_count` and `get_not_responding_peers` are currently used for other purposes as well, can we put an optional parameter to set custom expiry? For example,
```c++
void apply_to_not_responding_peers(const std::function<void(const ptr<peer>&)>& callback, int expiry = 0);
```
If given `expiry` is non-zero, honor the given `expiry`. If not, use `heart_beat_interval_ * reponse_limit_`.

antoniofilipovic:
Hi, @greensky00. Thanks for the fast response.

The change which you suggested makes sense. I will open the PR to fix it!

PR #523, opened 2024-07-24
Use leadership_expiry_ to deem nodes unhealthy

antoniofilipovic:
Opening PR as per issue https://github.com/eBay/NuRaft/issues/522

### DIFF
--- a/src/raft_server.cxx
+++ b/src/raft_server.cxx
@@ -633,7 +633,7 @@ void raft_server::apply_to_not_responding_peers(
     // default argument for expiry is used in case user defines leadership_expiry_.
     ptr<raft_params> params = ctx_->get_params();
 
-    if(expiry == 0){
+    if (expiry == 0) {
         expiry = params->heart_beat_interval_ * raft_server::raft_limits_.response_limit_;
     }
 
@@ -1120,7 +1120,7 @@ bool raft_server::check_leadership_validity() {
     int32 nr_peers{0};
 
     // Negative expiry: leadership will never expire.
-    if(leadership_expiry > 0){
+    if (leadership_expiry >= 0) {
         nr_peers = (int32)get_not_responding_peers_count(leadership_expiry);
     }
 
--- a/src/raft_server.cxx
+++ b/src/raft_server.cxx
@@ -602,13 +602,13 @@ int32 raft_server::get_leadership_expiry() {
     int expiry = params->leadership_expiry_;
     if (expiry == 0) {
         // If 0, default expiry: 20x of heartbeat.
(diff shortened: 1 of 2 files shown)


src/raft_server.cxx
```diff
@@ -617,21 +617,25 @@ std::list<ptr<peer>> raft_server::get_not_responding_peers() {
...
-    const std::function<void(const ptr<peer>&)>& callback) {
+    const std::function<void(const ptr<peer>&)>& callback, int expiry) {
     // Check if quorum nodes are not responding
-    // (i.e., don't respond 20x heartbeat time long).
+    // (i.e., don't respond 20x heartbeat time long or expiry if sent as argument).
+    // default argument for expiry is used in case user defines leadership_expiry_.
     ptr<raft_params> params = ctx_->get_params();
-    int expiry = params->heart_beat_interval_ * raft_server::raft_limits_.response_limit_;
+
+    if(expiry == 0){
```
greensky00, line 636: Please put space between terms:
```c++
if (expiry == 0) {
```

src/raft_server.cxx
```diff
@@ -1110,12 +1114,16 @@ bool raft_server::check_leadership_validity() {
...
     int leadership_expiry = get_leadership_expiry();
-    int32 nr_peers = (int32)get_not_responding_peers_count();
-    if (leadership_expiry < 0) {
-        // Negative expiry: leadership will never expire.
-        nr_peers = 0;
+
+    int32 nr_peers{0};
+
+    // Negative expiry: leadership will never expire.
+    if(leadership_expiry > 0){
```
greensky00, line 1123: Please put spaces too.
And the condition should be `if (leadership_expiry >= 0)`. `nr_peers` sets to `0` only when `leadership_expiry` is a negative number. If `leadership_expiry == 0`, it uses the default expiry (`response_limit_`).

greensky00 (commented):
Thanks, a couple of minor comments.

src/raft_server.cxx
antoniofilipovic: Thanks for the comment. Probably that is not the most clear part, but `get_leadership_expiry`() will return number greater than 0 if `params->leadership_expiry_` is set to 0.

greensky00 (approved):
Looks good, thanks.

src/raft_server.cxx
greensky00: Yeah, but this if condition needs to be correct on its own, without relying on the result of other function. Otherwise it will be mistake-prone and potentially introduce a bug silently.

PR #392, opened 2023-01-26
Fix data race in `asio_rpc_listener`

antonio2368:
TSAN in our sanitizer often caught a data race in `asio_rpc_listener`.

E.g. in
https://s3.amazonaws.com/clickhouse-test-reports/45430/7bb65cc002d27c325393f777773de6a1e1de9fdd/integration_tests__tsan__%5B2/6%5D.html
there is a [log](https://s3.amazonaws.com/clickhouse-test-reports/45430/7bb65cc002d27c325393f777773de6a1e1de9fdd/integration_tests__tsan__[2/6]/integration_run_parallel0_0.log) with
```
E           Exception: Sanitizer assert found in /ClickHouse/tests/integration/test_keeper_force_recovery_single_node/_instances_0/docker.log for instance ==================
```

Example of sanitizer warning:
```
WARNING: ThreadSanitizer: data race (pid=402)
  Write of size 8 at 0x7b4400017b08 by main thread (mutexes: write M0, write M1):
    #0 std::__1::enable_if<is_move_constructible<nuraft::raft_server*>::value && is_move_assignable<nuraft::raft_server*>::value, void>::type std::__1::swap[abi:v15000]<nuraft::raft_server*>(nuraft::raft_server*&, nuraft::raft_server*&) build_docker/../contrib/llvm-project/libcxx/include/__utility/swap.h:37:7 (clickhouse+0x22d909ef) (BuildId: a5753bde1234dfdac844bc70385bf039f8f9f009)
    #1 std::__1::shared_ptr<nuraft::raft_server>::swap[abi:v15000](std::__1::shared_ptr<nuraft::raft_server>&) build_docker/../contrib/llvm-project/libcxx/include/__memory/shared_ptr.h:761:9 (clickhouse+0x22d909ef)
    #2 std::__1::shared_ptr<nuraft::raft_server>::reset[abi:v15000]() build_docker/../contrib/llvm-project/libcxx/include/__memory/shared_ptr.h:768:22 (clickhouse+0x22d909ef)
    #3 nuraft::asio_rpc_listener::shutdown() build_docker/../contrib/NuRaft/src/asio_service.cxx:834:18 (clickhouse+0x22d909ef)
    #4 DB::KeeperServer::shutdownRaftServer() build_docker/../src/Coordination/KeeperServer.cpp:430:28 (clickhouse+0x1fc02dc7) (BuildId: a5753bde1234dfdac844bc70385bf039f8f9f009)
[...]

PR #524, opened 2024-07-26
Remove unnecessary listener_lock_ guard in start method

e0c9:


greensky00:
Hi, this lock is provided to prevent human error by ensuring that the caller of this function always acquires the lock, even though it doesn't actually do anything.

Please refer to this thread
https://github.com/eBay/NuRaft/pull/392#discussion_r1088612916

PR #527, opened 2024-08-13
asio_service support stream mode

ZZhongge:
1. set TCP_NODELAY = true this can be set after the socket open. So we can't set it in constructor. (async_connect will open the socket)
2. does the order of calling all the when_done matter when error occur? in write path when_done(error triggered) -> close_socket(in destructor) in read path close_socket(directly call) -> when_done(error triggered)
3. keep the old path for all request if stream_mode_ = false

already addressed all the comment.

### DIFF
--- a/src/asio_service.cxx
+++ b/src/asio_service.cxx
@@ -150,6 +150,31 @@ asio_service::meta_cb_params req_to_params(req_msg* req, resp_msg* resp) {
 
 // === ASIO Abstraction ===
 //     (to switch SSL <-> unsecure on-the-fly)
+struct pending_req_pkg {
+public:
+    pending_req_pkg(ptr<req_msg>& _req, 
+                    rpc_handler& _when_done, 
+                    uint64_t _timeout_ms = 0)
+        : req(_req), when_done(_when_done), timeout_ms(_timeout_ms)
+        {}
+
+    ptr<req_msg> get_req() {
+        return req;
+    }
+
+    rpc_handler& get_when_done() {
+        return when_done;
+    }
+
+    uint64_t get_timeout_ms() {
+        return timeout_ms;
+    }
+private:
+    ptr<req_msg> req;
+    rpc_handler when_done;
+    uint64_t timeout_ms;
+};
+
 class aa {
 public:
     template<typename BB, typename FF>
@@ -970,6 +995,8 @@ private:
                        const ERROR_CODE& err)
     {
         if (!err) {
+            asio::ip::tcp::no_delay option(true);
+            session->socket().set_option(option);
             p_in("receive a incoming rpc connection");
             session->prepare_handshake();
 
@@ -1124,13 +1151,38 @@ public:
             when_done(rsp, except);
             return;
         }
-        send(req, when_done, send_timeout_ms);
+        register_req_send(req, when_done, send_timeout_ms);
     }
 
     virtual void send(ptr<req_msg>& req,
                       rpc_handler& when_done,
                       uint64_t send_timeout_ms = 0) __override__
     {
+        if (impl_->get_options().streaming_mode_) {
+            pre_send(req, when_done, send_timeout_ms);
+        } else {
+            register_req_send(req, when_done, send_timeout_ms);
+        }
+    }
+
+    void pre_send(ptr<req_msg>& req,
+                    rpc_handler& when_done,
+                    uint64_t send_timeout_ms) {
+        auto_lock(pending_write_reqs_lock_);
+        if (pending_write == 0) {
+            register_req_send(req, when_done, send_timeout_ms);
+        } else {
+            pending_write_reqs_.push_back(cs_new<pending_req_pkg>(req, when_done, send_timeout_ms));
+        }
+
+        pending_write++;
+        p_db("start to send msg to peer %d, start_log_idx: %ld, size: %ld, pending write reqs: %ld", 
+        req->get_dst(), req->get_last_log_idx(), req->log_entries().size(), pending_write);
+    }
+
+    void register_req_send(ptr<req_msg>& req,
+                      rpc_handler& when_done,
+                      uint64_t send_timeout_ms) {
         if (abandoned_) {
             p_er( "client %p to %s:%s is already stale (SSL %s)",
                   this, host_.c_str(), port_.c_str(),
@@ -1447,6 +1499,13 @@ private:
         } );
     }
 
+    void free_busy_flag(bool by_write) {
+        // Either satisfied or neither satisfied
+        if (impl_->get_options().streaming_mode_ == by_write) {
+            set_busy_flag(false);
+        }
+    }
+
     void set_busy_flag(bool to) {
         if (to == true) {
             bool exp = false;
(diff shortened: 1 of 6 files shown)


tests/CMakeLists.txt
```diff
@@ -32,6 +32,17 @@ target_link_libraries(asio_service_test
...
+add_executable(asio_service_stream_test
+                unit/asio_service_stream_test.cxx
+                ${EXAMPLES_SRC}/logger.cc
+                ${EXAMPLES_SRC}/in_memory_log_store.cxx)
+add_dependencies(asio_service_stream_test
+                static_lib
+                build_ssl_key)
+target_link_libraries(asio_service_stream_test
+                ${BUILD_DIR}/${LIBRARY_OUTPUT_NAME}
+                ${LIBRARIES})
```
greensky00, line 44: Indent.

tests/unit/asio_service_stream_test.cxx
```diff
@@ -0,0 +1,227 @@
...
+        int my_id;
+        int port;
+        std::atomic<ulong> response_log_index;
+        ulong send_log_index = 0;
+        ptr<asio_service> asio_svc;
+        ptr<rpc_client> my_client;
+        ptr<rpc_listener> my_listener;
+        ptr<logger_wrapper> my_log_wrapper;
+        ptr<logger> my_log;
+        ptr<stream_msg_handler> my_msg_handler;
```
greensky00, line 161: Use `_` suffix for all member variable names.

tests/unit/asio_service_stream_test.cxx
```diff
@@ -0,0 +1,227 @@
...
+            }
+        }
+
+        void handle_result(ptr<req_msg>& req, ptr<resp_msg>& resp, ptr<rpc_exception>& err) {
+            if (resp->get_next_idx() == get_next_log_index()) {
+                response_log_index++;
+            } else {
+                SimpleLogger* ll = my_log_wrapper->getLogger();
+                _log_info(ll, "resp log index not match, resp: %ld, current: %ld", 
+                resp->get_next_idx(), get_next_log_index());
```
greensky00, line 118: This is the whole point of this test (checking whether request and response are delivered in order). Rather than leaving a log, there should be a boolean flag to set if failure happens, and check `CHK_FALSE` at the end.

tests/unit/asio_service_stream_test.cxx
```diff
@@ -0,0 +1,227 @@
...
+                std::string buf_str = buf->get_str();
+                if (buf_str != test_msg) {
+                    SimpleLogger* ll = my_log_wrapper->getLogger();
+                    _log_info(ll, "resp str: %s", buf_str.c_str());
+                    msg_mismatch.store(true);
+                }
+            } else {
+                SimpleLogger* ll = my_log_wrapper->getLogger();
+                _log_info(ll, "req log index not match, req: %ld, current: %ld", 
+                req.get_last_log_idx(), streamed_log_index.load());
```
greensky00, line 66: Ditto, it should set a boolean flag and make test failed.

greensky00:
Also, can you add
```
  pull_request:
    branches:
      - master
      - streaming
```
into `.github/workflow/cmake.yml`? in order to trigger PR validation for this branch.

src/asio_service.cxx
```diff
@@ -1124,13 +1151,38 @@ class asio_rpc_client
...
+        } else {
+            register_req_send(req, when_done, send_timeout_ms);
+        }
+    }
+
+    void pre_send(ptr<req_msg>& req,
+                    rpc_handler& when_done,
+                    uint64_t send_timeout_ms) {
+        auto_lock(pending_write_reqs_lock_);
+        if (pending_write == 0) {
```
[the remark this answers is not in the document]
ZZhongge: When the queue is empty, the coming request will be sent immediately and not be added to this queue. So when the queue is empty there are two cases. One is no flying request and another is we have one flying request.

tests/unit/asio_service_stream_test.cxx
ZZhongge: If entering into this branch, this `response_log_index` will not increase. So it will fail at `CHK_EQ(count, s.get_resp_log_index());`. Those logs is just for helping debug.

src/asio_service.cxx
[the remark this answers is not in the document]
ZZhongge: In close_socket, the when_done of the request that triggered the error have already been invoked. So to avoid call the when_done twice. Request that being processed is not stored in the queue.

src/asio_service.cxx
```diff
@@ -1481,6 +1540,42 @@ class asio_rpc_client
...
 #endif
+        if (!impl_->get_options().streaming_mode_) {
+            return;
+        }
+        // clear write queue and read queue here
+        // to keep the order, latest to oldest (is that ok?)
+        // handle write queue first in reverse
+        {
+            auto_lock(pending_write_reqs_lock_);
+            for (auto rit = pending_write_reqs_.rbegin(); rit != pending_write_reqs_.rend(); ++rit) {
```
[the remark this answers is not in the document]
ZZhongge: The previous code will call close_socket first. Shall we need to change the order of calling `when_done` and `close_socket`
```
            operation_timer_.cancel();
            abandoned_ = true;
            ptr<resp_msg> rsp;
            ptr<rpc_exception> except
                ( cs_new<rpc_exception>
                  ( sstrfmt( "failed to send request to peer %d, %s:%s, "
                             "error %d, %s" )
                           .fmt( req->get_dst(), host_.c_str(),
                                 port_.c_str(), err.value(), err.message().c_str() ),
                    req ) );
            close_socket();
            when_done(rsp, except);
```

src/asio_service.cxx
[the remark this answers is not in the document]
greensky00: I understand your point about this variable, but maintaining a separate counter (which has different value from `pending_write_reqs_.size()`) will be very confusing and mistake-prone during the future maintenance.

Let's do this way; put the current in-flight request into the queue as well:
```
    void pre_send(ptr<req_msg>& req,
                  rpc_handler& when_done,
                  uint64_t send_timeout_ms) {
        auto_lock(pending_write_reqs_lock_);
        pending_write_reqs_.push_back(
            cs_new<pending_req_pkg>(req, when_done, send_timeout_ms));
        if (pending_write_reqs_.size() == 1) {
            register_req_send(req, when_done, send_timeout_ms);
        }
        ...
    }
```
and
```
    void post_send(ptr<req_msg>& req, rpc_handler& when_done) {
	    ...
        // next process write
        {
            auto_lock(pending_write_reqs_lock_);
            pending_write_reqs_.pop_front();
            if (pending_write_reqs_.size() > 0) {
                ptr<pending_req_pkg> next_req_pkg = *pending_write_reqs_.begin();
                register_req_send(next_req_pkg->req_, next_req_pkg->when_done_, 
                                  next_req_pkg->timeout_ms_);                
                p_db("trigger next write, start_log_idx: %ld, pending write reqs: %ld",
                     next_req_pkg->req_->get_last_log_idx(), pending_write_);
            }
        }
    }
```
same to read requests as well. This also resolves the issue with `when_done()` and `close_socket()`; when error happens, if `streaming_mode_` is on, we just call `close_socket()` without calling `when_done()`.

greensky00 (approved):

PR #530, opened 2024-08-22
Implement stream mode in peer and raft server(a proposal)

ZZhongge:
add an additional flag: flying_append_entry_request.
  it is used to present if it is the first request. if streaming is disable in the middle, we can only send the new request when flying request = 0

### DIFF
--- a/include/libnuraft/peer.hxx
+++ b/include/libnuraft/peer.hxx
@@ -77,6 +77,9 @@ public:
         , lost_by_leader_(false)
         , rsv_msg_(nullptr)
         , rsv_msg_handler_(nullptr)
+        , last_streamed_log_idx_(0)
+        , max_log_gap_in_stream_( ctx.get_params()->max_log_gap_in_stream_ )
+        , flying_append_entry_request_(0)
         , l_(logger)
     {
         reset_ls_timer();
@@ -303,6 +306,37 @@ public:
     ptr<req_msg> get_rsv_msg() const { return rsv_msg_; }
     rpc_handler get_rsv_msg_handler() const { return rsv_msg_handler_; }
 
+    ulong get_last_streamed_log_idx() {
+        return last_streamed_log_idx_.load();
+    }
+
+    void set_last_streamed_log_idx(ulong expected, ulong idx) {
+        last_streamed_log_idx_.compare_exchange_strong(expected, idx);
+    }
+
+    bool is_streaming() {
+        return max_log_gap_in_stream_ > 0;
+    }
+
+    int32 get_max_log_gap_in_stream() {
+        return max_log_gap_in_stream_;
+    }
+
+    void reset_stream() {
+        last_streamed_log_idx_.store(0);
+    }
+
+    bool start_append_entry() {
+        ulong expected = 0;
+        return flying_append_entry_request_.compare_exchange_strong(expected, 1);
+    }
+
+    void add_append_entry_request() {
+        flying_append_entry_request_.fetch_add(1);
+    }
+
+    void try_set_free(msg_type type);
+
     bool is_lost() const { return lost_by_leader_; }
     void set_lost() { lost_by_leader_ = true; }
     void set_recovered() { lost_by_leader_ = false; }
@@ -518,6 +552,18 @@ private:
      * Handler for reserved message.
      */
     rpc_handler rsv_msg_handler_;
+
+    /**
+     * Last log index sent in stream mode.
+     */
+    std::atomic<ulong> last_streamed_log_idx_;
+
+    int32 max_log_gap_in_stream_;
+
+    /**
+     * if `true`, this peer is in stream mode.
+     */
+    std::atomic<ulong> flying_append_entry_request_;
 
     /**
      * Logger instance.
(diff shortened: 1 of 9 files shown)


src/peer.cxx
```diff
@@ -231,6 +240,8 @@ bool peer::recreate_rpc(ptr<srv_config>& config,
         //   hence reset timer.
         reset_active_timer();
 
+        reset_stream();
```
greensky00, line 243: This should also be called in `handle_rpc_result`, when `err` exists.

include/libnuraft/peer.hxx
```diff
@@ -77,6 +77,9 @@ public:
         , lost_by_leader_(false)
         , rsv_msg_(nullptr)
         , rsv_msg_handler_(nullptr)
+        , last_streamed_log_idx_(0)
+        , max_log_gap_in_stream_( ctx.get_params()->max_log_gap_in_stream_ )
```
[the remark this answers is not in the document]
ZZhongge: I make a local copy here because in peer we can't access ctx. So we can't determine if we need to set_free.

src/handle_append_entries.cxx
```diff
@@ -398,7 +446,11 @@ ptr<req_msg> raft_server::create_append_entries_req(ptr<peer>& pp) {
             p.set_next_log_idx(cur_nxt_idx);
         }
 
-        last_log_idx = p.get_next_log_idx() - 1;
+        if (last_streamed_log_idx > 0) {
+            last_log_idx = last_streamed_log_idx;
+        } else {
+            last_log_idx = p.get_next_log_idx() - 1;
+        }
```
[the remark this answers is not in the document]
ZZhongge: I use the number here. Because I want to make sure that the **last_streamed_log_idx** won't change during I process the request. When we enter into this function with stream = true, the response may set it to 0 in the middle of processing. Sometimes, the last_log_idx may be 0.

src/peer.cxx
ZZhongge: How can we handle the rejected request in stream mode? It is not an error.

ZZhongge: I see that we call set_free() here. Don't we need to call reset_stream() when the rpc recreated?

src/handle_append_entries.cxx
```diff
@@ -271,88 +271,78 @@ bool raft_server::request_append_entries(ptr<peer> p) {
...
             }
-            p->reset_long_pause_warnings();
-
-        } else {
-            // FIXME: `manual_free` is deprecated, need to get rid of it.
-
-            // It means that this is not an actual recovery,
-            // but just temporarily freed busy flag.
-            p->reset_manual_free();
+            return send_request(p, msg, m_handler);
```
greensky00, line 344: Need refactoring due to duplicate code. I'd rather do
```c++
    ptr<req_msg> msg = p->get_rsv_msg();
    rpc_handler m_handler = p->get_rsv_msg_handler();
    if (msg) {
        if (p->make_busy()) {
            // Clear the reserved message.
            p->set_rsv_msg(nullptr, nullptr);
            p_in("found reserved message to peer %d, type %d",
                    p->get_id(), msg->get_type());
            return send_request(p, msg, m_handler);
        }
    } else {
        ulong last_streamed_log_idx = p->get_last_streamed_log_idx();
        bool streaming = last_streamed_log_idx > 0;

        if (streaming || p->make_busy()) {
            msg = create_append_entries_req(p, last_streamed_log_idx);
            m_handler = resp_handler_;

            if (msg) {
                streaming = streaming && 
                            msg->get_type() == msg_type::append_entries_request;
                bool make_busy_result = p->is_busy();

                if (streaming) {
                    p->set_last_streamed_log_idx(
                        last_streamed_log_idx, 
                        last_streamed_log_idx + msg->log_entries().size());
                } else if (!make_busy_result) {
                    make_busy_result = p->make_busy();
                }

                if (streaming || make_busy_result) {
                    return send_request(p, msg, m_handler, streaming);
                }
            } else {
                if (!streaming) {
                    p->set_free();
                }
                if (params->use_bg_thread_for_snapshot_io_ &&
                    p->get_snapshot_sync_ctx() ) {
                    // If this is an async snapshot request, invoke IO thread.
                    snapshot_io_mgr::instance().invoke();
                }
                return true;
            }
        }
    }
```

src/handle_append_entries.cxx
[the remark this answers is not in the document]
greensky00: Then it will be better to give a generic name `custom_last_log_idx`:
```c++
ptr<req_msg> raft_server::create_append_entries_req(ptr<peer>& pp,
                                                    ulong custom_last_log_idx)
```
and move `p.set_last_streamed_log_idx()` out of `create_append_entries_req`, as I put in the above comment. `create_append_entries_req` is not a proper place to update it (as it is unaware of streaming).

greensky00 (approved):

tests/CMakeLists.txt
```diff
@@ -43,6 +43,17 @@ target_link_libraries(asio_service_stream_test
                       ${BUILD_DIR}/${LIBRARY_OUTPUT_NAME}
                       ${LIBRARIES})
 
+add_executable(asio_stream_test
+               unit/asio_stream_test.cxx
```
greensky00, line 47: By the way, in the next PR (for testing), please rename it to `stream_functional_test`.

Issue #535, opened 2024-09-12
Use `pthread_exit` instead of `exit` to allow proper process shutdown.


szmyd:
When NuRaft encounters an unrecoverable state it calls `exit(2)` directly, such as:
```c++
 // handle_commit.cxx
 160 bool raft_server::commit_in_bg_exec(size_t timeout_ms) {
 ...
 225         ptr<log_entry> le = log_store_->entry_at(index_to_commit);
 226         if (!le)
 227         {
 228             // LCOV_EXCL_START
 229             p_ft( "failed to get log entry with idx %" PRIu64 "", index_to_commit );
 230             ctx_->state_mgr_->system_exit(raft_err::N19_bad_log_idx_for_term);
 231             ::exit(-1);
 232             // LCOV_EXCL_STOP
 233         }
```
`commit_bg_exec` runs in a separate thread from the `main` process which could also have other threads or even other nuraft::raft_server's running. This forceful termination does not allow the process to try and enter a "non-random" state and behaves as if every thread abnormally terminated.

Requested is that `pthread_exit(3)` be used instead after a call to `state_mgr::system_exit` to announce that the server is going to stop processing. The main process can decide whether to also terminate and `join` all other threads normally, continue after abandoning this raft service or abend itself.

PR #536, opened 2024-09-12
Use pthread_exit to terminate raft service.

szmyd:
Allow the process to handle termination of raft service rather than force process wide `exit()`

Addresses #535

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,6 +78,7 @@ endif()
 
 
 # === Compiler flags ===
+option(USE_PTHREAD_EXIT "Call pthread_exit on server threads" OFF)
 if (NOT WIN32)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall")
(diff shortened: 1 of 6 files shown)


szmyd:
I dunno if this should have a compile time condition around it similar to:
```c++
#ifdef USE_PTHREAD_EXIT
pthread_exit(nullptr);
#else 
exit(-1)
#endif
```
to avoid behavior change for those relying on this?

greensky00:
Does it work correctly on Mac? I believe it doesn't work at least on Windows. We should write a helper function that works differently according to compiler flag, as you mentioned.

szmyd:
> Does it work correctly on Mac? I believe it doesn't work at least on Windows. We should write a helper function that works differently according to compiler flag, as you mentioned.

It definitely doesn't work on Windows; but _should_ on Mac...Mac has `pthread_exit(3)`. I guess the question is, if it's compile time determined what should be the default? `exit()`?

greensky00 (approved):
Looks good, let's revisit later.

PR #561, opened 2025-01-10
Add timeout for snapshot_sync_ctx

yuwmao:
Add a dedicated timeout for snapshot_sync_ctx, and user can set this value in raft_param.

### DIFF
--- a/include/libnuraft/raft_params.hxx
+++ b/include/libnuraft/raft_params.hxx
@@ -74,6 +74,7 @@ struct raft_params {
         , log_sync_stop_gap_(99999)
         , snapshot_distance_(0)
         , snapshot_block_size_(0)
+        , snapshot_sync_ctx_timeout_(0)
         , enable_randomized_snapshot_creation_(false)
         , max_append_size_(100)
         , reserved_log_items_(100000)
@@ -218,6 +219,17 @@ struct raft_params {
      */
     raft_params& with_snapshot_sync_block_size(int32 size) {
         snapshot_block_size_ = size;
+        return *this;
+    }
+
+    /**
+    * Timeout for syncing the snapshot requests.
+    *
+    * @param timeout_ms
+    * @return self
+    */
+    raft_params& with_snapshot_sync_ctx_timeout(int32 timeout_ms) {
+        snapshot_sync_ctx_timeout_ = timeout_ms;
         return *this;
     }
 
@@ -404,6 +416,13 @@ public:
      * (Deprecated).
      */
     int32 snapshot_block_size_;
+
+    /**
+        * Timeout(ms) for snapshot_sync_ctx, if a single snapshot syncing request exceeds this,
+        * it will be considered as timeout and ctx will be released.
+        * 0 means it will be set to the default value `heart_beat_interval_ * response_limit_`.
+        */
+    int32 snapshot_sync_ctx_timeout_;
 
     /**
      * Enable randomized snapshot creation which will avoid
--- a/include/libnuraft/raft_params.hxx
+++ b/include/libnuraft/raft_params.hxx
@@ -223,11 +223,11 @@ struct raft_params {
     }
 
     /**
-    * Timeout for syncing the snapshot requests.
-    *
-    * @param timeout_ms
-    * @return self
-    */
+     * Timeout for syncing the snapshot requests.
+     *
+     * @param timeout_ms
+     * @return self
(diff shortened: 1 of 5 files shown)


src/handle_timeout.cxx
```diff
@@ -79,8 +79,7 @@ void raft_server::handle_hb_timeout(int32 srv_id) {
 
             ptr<raft_params> params = ctx_->get_params();
             uint64_t resp_timer_ms = srv_to_join_->get_resp_timer_us() / 1000;
-            if ( resp_timer_ms >= (uint64_t)params->heart_beat_interval_ *
-                                  raft_server::raft_limits_.response_limit_ ) {
+            if (resp_timer_ms >= (uint64_t)params->snapshot_sync_ctx_timeout_) {
```
xiaoxichen, line 82: if we use bg thread for snapshot, will we still get HB timeout here?

src/handle_join_leave.cxx
```diff
@@ -91,9 +91,7 @@ ptr<resp_msg> raft_server::handle_add_srv_req(req_msg& req) {
              srv_to_join_->get_id(),
              last_active_ms);
 
-        if ( last_active_ms <=
-                 (ulong)raft_server::raft_limits_.response_limit_ *
-                 ctx_->get_params()->heart_beat_interval_ ) {
+        if (last_active_ms <= (ulong)ctx_->get_params()->snapshot_sync_ctx_timeout_) {
```
xiaoxichen, line 94: can we take a 
```
max ((ulong)raft_server::raft_limits_.response_limit_ *
                 ctx_->get_params()->heart_beat_interval_ , (ulong)ctx_->get_params()->snapshot_sync_ctx_timeout_)
                 ```
also considering rename  `snapshot_sync_ctx_timeout_` to more general initial_sync_timeout?  as this path is not snapshot only.

File: include/libnuraft/raft_params.hxx [hunk 3]
```diff
@@ -74,6 +74,7 @@ struct raft_params {
         , log_sync_stop_gap_(99999)
         , snapshot_distance_(0)
         , snapshot_block_size_(0)
+        , snapshot_sync_ctx_timeout_(2500)
```
greensky00, line 77: Put `0` by default, and if zero, keep the current behavior (`heart_beat_interval_ * response_limit_`). There are lots of users with different heartbeats, and such a newly introduced parameter with a fixed default value will suddenly break their systems.

src/handle_join_leave.cxx
yuwmao: Make sense, will take into account `response_limit_ * heart_beat_interval_` here.

src/handle_timeout.cxx
yuwmao: Good point. I think so. In my understanding, heartbeat timeout and snapshot timeout are two things.
This HB timeout logic is kept in `srv_to_join_` [as a handler](https://github.com/eBay/NuRaft/blob/cd4ff77592669ebb0aac905cd3ef3144345a7531/src/handle_join_leave.cxx#L111) and seems HB is used for [retry in `use_bg_thread_for_snapshot_io_` mode](https://github.com/eBay/NuRaft/blob/2c2d2c23447701517677de9bfd3aeaa896ee7cd4/src/snapshot_sync_ctx.cxx#L214).

And for the normal snapshot path in `use_bg_thread_for_snapshot_io_` mode, it uses the same [ex_resp_handler_](https://github.com/eBay/NuRaft/blob/cd4ff77592669ebb0aac905cd3ef3144345a7531/src/handle_snapshot_sync.cxx#L165) and snapshot timeout will be checked in next `create_sync_snapshot_req`([link](https://github.com/eBay/NuRaft/blob/cd4ff77592669ebb0aac905cd3ef3144345a7531/src/handle_snapshot_sync.cxx#L96)).

src/handle_join_leave.cxx
yuwmao: I think `initial_sync_timeout` is the result of `max(response_limit_ * heart_beat_interval_, snapshot_sync_ctx_timeout_)`, right?
How about using `srv_to_join_->get_snapshot_sync_ctx()` to see if it is in snapshot process or not?

xiaoxichen (approved):
lgtm

greensky00 (approved):
Thanks, looks good. Fixed indent and put a comment.

PR #571, opened 2025-03-10
Support learner change

yuwmao:
Previously, learner is a static srv config, provide a cmd for use to flip the learner flag.

### DIFF
--- a/include/libnuraft/raft_server.hxx
+++ b/include/libnuraft/raft_server.hxx
@@ -261,6 +261,18 @@ public:
         append_entries(const std::vector< ptr<buffer> >& logs);
 
     /**
+     * Flip learner flag of given server.
+     * Learner will be excluded from the quorum.
+     * Only leader will accept this operation.
+     * This is also an asynchronous task.
+     *
+     * @param srv_id ID of the server to set as a learner.
+     * @param to If `true`, set the server as a learner, otherwise, clear learner flag.
+     * @return `ret->get_result_code()` will be OK on success.
+     */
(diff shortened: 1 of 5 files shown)


include/libnuraft/msg_type.hxx
```diff
@@ -60,6 +60,8 @@ enum msg_type {
     reconnect_response              = 27,
     custom_notification_request     = 28,
     custom_notification_response    = 29,
+    learner_change_request          = 30,
+    learner_change_response         = 31,
```
greensky00, line 64: New message types are not needed.

Same as priority change (`raft_server::set_priority`), we can just create a new config log with updated server config, and replicate it. There is no need for its handler.
https://github.com/eBay/NuRaft/blob/master/src/handle_priority.cxx#L79-L106

There is a special message type `priority_change_request`, in order to broadcast (temporary) priority change in case when the leader is dead, for the sake of troubleshooting. This shouldn't be the case for `learner` flag, as changing `learner` role without leader is not safe.

yuwmao: I saw there is an auto_forwarding_ parameter, and when enabled, add_srv/remove_srv can be forwarded to leader. Not sure in this scenario if we should allow forwarding for learner change.

yuwmao:
Remove `learner_change_request`. 
Use cmd_result as return value to reuse error code.

xiaoxichen (commented):
lgtm

greensky00 (approved):
Thanks, fixed some indent.

PR #585, opened 2025-04-02
Problem: can't use system OpenSSL

yrashk:
When configured to use system OpenSSL, OpenSSL::SSL and OpenSSL::Crypto targets are attempted to link.

Solution: use actual libraries instead of targets


greensky00:
Hi @yrashk 
Do you see actual build failure due to this? Both `OpenSSL::SSL` and `OpenSSL::Crypto` are set as a result of `FindOpenSSL`. If `FindOpenSSL` succeeds (hence `OPENSSL_FOUND` is set), the above both should work as well.
https://cmake.org/cmake/help/v3.26/module/FindOpenSSL.html

yrashk:
Yes, I do see an actual build failure in my case. I have a bit of a different scenario, though. I have infrastructure that does its own passes for finding OpenSSL, but it does use `find_package(OpenSSL)` underneath, before it proceeds with NuRaft.

I'll try to find a different solution to my problem locally, then.

greensky00:
I'm not sure I understand your case correctly, but you can set `-DDEPS_PREFIX=<path>` or `-DOPENSSL_ROOT_DIR=<path>` if you have your own openssl path.

yrashk:
I suspect I have a better idea of what happened. I used a slightly older version of NuRaft (shortly before v3) and continued on with the updated one but unchanged build directory. That's where I saw the failure to find these targets.

Once I wiped out the build directory and re-ran `cmake -S -B` to re-create it, everything went fine.

greensky00:
Got it, so is it ok to close this PR?

yrashk:
Yep!

PR #587, opened 2025-04-08
Fix changing busy connection limit

antonio2368:
`set_raft_limits` uses assignment operator so you couldn't change busy limit with it.


greensky00 (approved):
Thanks!

PR #656, opened 2026-06-22
Expose per-peer committed log index in peer_info

xiaoxichen:
Add `last_committed_idx_` to `raft_server::peer_info` so callers can detect peers that have appended logs but cannot commit them (e.g., due to deadlocked commit threads). The value is populated from the existing `peer::get_sm_committed_idx()` and is meaningful only when `raft_params::track_peers_sm_commit_idx_` is enabled.

### DIFF
--- a/include/libnuraft/raft_server.hxx
+++ b/include/libnuraft/raft_server.hxx
@@ -653,6 +653,7 @@ public:
         peer_info()
             : id_(-1)
             , last_log_idx_(0)
+            , last_committed_idx_(0)
             , last_succ_resp_us_(0)
             {}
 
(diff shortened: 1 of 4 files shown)


include/libnuraft/raft_server.hxx
```diff
@@ -666,6 +667,18 @@ public:
...
+         * The last committed log index of the peer's state machine,
+         * from this server's point of view. It can be smaller than
+         * `last_log_idx_` if the peer has appended the log but has not
+         * committed it yet (e.g., due to slow or deadlocked commit threads).
+         *
+         * NOTE: This field is meaningful only when
+         * `raft_params::track_peers_sm_commit_idx_` is enabled. Otherwise,
+         * it will always be 0.
+         */
+        ulong last_committed_idx_;
```
greensky00: Please rename it to `last_sm_committed_idx_`.

Otherwise, it is confusing whether this refers to the most recent committed log index in the log store or the log index that has already been applied to the state machine (<= this is what you want).

greensky00 (commented):
Looks good, only renaming.

greensky00 (approved):

PR #657, opened 2026-06-25
Fix snapshot retry rewind handling

Besroy:
When a follower reports RECEIVING_SNAPSHOT, the leader should retry snapshot sync instead of applying the normal append failure rewind. Rewinding next_log_idx from 1 to 0 causes the next request path to treat the peer progress as uninitialized and reset it to the leader's current next_slot, which can make create_sync_snapshot_req validate against an impossible last_log_idx.

Keep next_log_idx aligned with the follower response and skip the rewind for RECEIVING_SNAPSHOT so snapshot retry starts from the expected peer progress.

### DIFF
--- a/src/handle_append_entries.cxx
+++ b/src/handle_append_entries.cxx
@@ -1345,6 +1345,10 @@ void raft_server::handle_append_entries_resp(resp_msg& resp) {
                     do_log_rewind = false;
                 } else if (appendix->extra_order_ == resp_appendix::RECEIVING_SNAPSHOT) {
                     p->set_snapshot_sync_is_needed(true);
+                    if (resp.get_next_idx() > 0) {
+                        p->set_next_log_idx(resp.get_next_idx());
+                    }
+                    do_log_rewind = false;
                     p_in("peer %d was in snapshot sync mode, re-sending a snapshot. "
                          "peers next log idx: %" PRIu64 ", resp next idx: %" PRIu64,
                          p->get_id(), prev_next_log, resp.get_next_idx());
--- a/tests/asio/custom_quorum_test.cxx
+++ b/tests/asio/custom_quorum_test.cxx
@@ -1405,6 +1405,7 @@ int full_consensus_with_snapshot_transfer_test() {
     restart_params.with_election_timeout_upper(RaftAsioPkg::HEARTBEAT_MS * 4);
     restart_params.with_reserved_log_items(5);
     restart_params.with_snapshot_enabled(10);
+    restart_params.with_max_append_size(5);
     restart_params.with_client_req_timeout(10000);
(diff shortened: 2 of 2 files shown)


src/raft_server.cxx
```diff
@@ -675,10 +675,11 @@ bool raft_server::is_excluded_from_quorum(const peer& pp,
     if (resp_elapsed_ms <= expiry) {
         // Response time is within the expiry time.
 
-        if (required_log_idx &&
-            ((pp.get_matched_idx() &&
-              pp.get_matched_idx() < required_log_idx) ||
-             pp.get_snapshot_sync_ctx().get())) {
+        if (pp.is_snapshot_sync_needed() ||
```
xiaoxichen: Although I think it is right but i didnt get the context of this change

Besroy: This change is to handle two corner cases:
1. The leader has received the RECEIVING_SNAPSHOT flag and set snapshot_sync_needed, but the new snapshot sync context has not been created yet.
2. required_log_idx can be 0, for example when quick_commit_index_ <= max_append_size_. In that case, gating the snapshot-sync check behind required_log_idx would skip the exclusion logic entirely.

greensky00: `required_log_idx` is given only when full consensus mode is on, to get the minimum sm commit index among members in full consensus group, by excluding members who is confirmed as 1) self mark down, 2) marked by leader, 3) leader is sending snapshot. So `pp.is_snapshot_sync_needed()` is not acceptable, as it is just a request from follower, but not leader's official confirm (`pp.get_snapshot_sync_ctx()`). This is very strict condition related to correctness.

Also, for normal case (`required_log_idx == 0`), this is used for health check of followers, and followers receiving snapshot and responding in time is treated as healthy, so they should not be excluded. Otherwise, a valid leader will end up with incorrect resignation.

Besroy: Thanks for the feedback. Changes reverted.                                                                                                                                                                                                                                                                                            

The original motivation was a UT failure in full_consensus_with_snapshot_transfer_test. The root cause was that quick_commit_index_ (29 in the test) was smaller than max_append_size_ (default 100), which caused allowed_log_index to be computed as 0. With required_log_idx == 0, the get_snapshot_sync_ctx() check in is_excluded_from_quorum is skipped, so S3 (receiving snapshot) was not excluded from the full consensus quorum, and the new entry couldn't be committed within the test's wait window.                                                                                                                                                                                                                         

In production, snapshot_distance is typically larger than max_append_size, which means quick_commit_index_ will already exceed max_append_size by the time a snapshot exists to send — so allowed_log_index will be non-zero and the exclusion logic works correctly. Given this, I've fixed the UT instead.

greensky00 (approved):

